PR #4, opened 2024-07-16
setup a basic test infrastructure

zerbina:
## Summary

Add the `tester` tool, which is a simple test driver handling test
discovery and output checks. The actual test execution is handled on a
per-category basis via separate `runner` programs.

## Details

* add the `tester` tool and register it with `koch`
* run the tester as part of CI
* add a basic runner and test to make sure the tester works

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -62,3 +62,6 @@ jobs:
 
       - name: Build all programs
         run: bin/koch all -d:nimStrictMode -d:release
+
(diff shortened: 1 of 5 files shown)


zerbina:
@saem: The `tester` is really only meant to get the test infrastructure off the ground. I think the idea with the separate `runner` programs is solid, as it allows for quick iteration and low-cost experimentation, but don't feel obligated to keep the current `tester` code around when improving the test infrastructure.

saem (approved):

PR #6, opened 2024-07-19
document the planned nanopass architecture

zerbina:
The document is intended to capture the current plans and ideas and act
as a reference.

### DIFF
--- a/passes/design.md
+++ b/passes/design.md
@@ -39,7 +39,7 @@ Top ::= ('.extends' Identifier)? (Def | Append | Remove)*
 
 A `Def` defines a named rule, an `Append` appends to an existing named rule,
 and a `Remove` removes the given rule(s) from the named rule. A `Reference`
-is a reference to another name rule. There are two special references: `<int>`
+is a reference to another named rule. There are two special references: `<int>`
 and `<float>`. They refer to a literal integer and float value, respectively.
 
 The idea is to use S-expressions as the serialization format (it's both human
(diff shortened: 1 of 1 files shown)


passes/design.md
```diff
@@ -0,0 +1,47 @@
...
+Def ::= Identifier '::=' TopMatcher ('|' TopMatcher)*
+Append ::= Identifier '+=' TopMatcher ('|' TopMatcher)*
+Remove ::= Identifier '-=' TopMatcher ('|' TopMatcher)*
+
+Top ::= ('.extends' Identifier)? (Def | Append | Remove)*
+```
+
+A `Def` defines a named rule, an `Append` appends to an existing named rule,
+and a `Remove` removes the given rule(s) from the named rule. A `Reference`
+is a reference to another name rule. There are two special references: `<int>`
```
saem, line 42:
```suggestion
is a reference to another named rule. There are two special references: `<int>`
```

passes/design.md
```diff
@@ -0,0 +1,47 @@
+## Nanopass Architecture
+
+The idea is to structure compilation as repeatedly transforming the source
+language into increasingly lower-level, well-defined intermediate languages
```
saem, line 4: Just a note: Technically we can also go back up via a pass designed to do as such. So given language `b`, which is higher level than language `a`, we could have one pass to lower from `b` to `a`, and another to raise from `a` to `b`.

I don't think we need to document this just yet, at least not until we run into an example of the need in the wild, and then we can speak to the conditions under which we might want to do it.

saem (approved):
Just one possible typo.

Also, left a note about using passes for raising but I think documenting at this point might be more academic than useful. It's there so we, or others, can reference it in the future if the need arises.

PR #7, opened 2024-07-21
implement first iteration of the lowest-level language

zerbina:
## Summary

* add a specification, pass, test runner, and tests for the lowest-
  level language that immediately comes before VM bytecode
* add a generic `PackedTree` implementation for use as the passes' IR

### Lanuage Design

* the language provides simple abstractions over some VM concepts
* all global entity descriptions (types, globals, procs) are stored in
  separate sections, and they're referenced by index (for fast lookup)
* local continuations abstract over raw jumps
* the type system makes a distinction between signed and unsigned
  integers

In its current form, the language is likely too high-level, but it's a
good start to base further language development on.

### Pass Design

The pass currently operates on whole modules at a time (though there's
no multi-module support at the moment). Beyond some minor assertions,
syntax and proper typing is *not* checked by the pass -- the idea is
that during normal compilation, passes trust that their input is sane
and sound, with validation implemented separately.

### Storage

A simple generic `PackedTree` type (based on NimSkull's `MirTree`) is
used as the IR for the pass. The node kind enum to use is provided by
the `spec` module, alongside S-expression serialization/
deserialization.

Nodes only store the kind and a type-erased value. Extra information,
such as types, needs to use nodes. This keeps nodes small (currently
8-byte) and serialization easy.

---

## To-Do
- [x] add the remaining tests


zerbina:
For comparison, a `MirNode` currently has a size of 16 bytes. 8 byte is a good size, because it
means that a node fits into a single register (on a 64-bit architecture).

## Next Step

The next step will be implementing a tool that comprehends the grammar from the Markdown files
(I'm already using an early implementation thereof locally). My plan is that it is responsible
for:
* making sure the grammar descriptions are syntactically valid
* making sure the grammar descriptions are valid (e.g., no ambiguities, terminates, all used
  names exists, etc.)
* generating the `spec` module based on the (merged) grammars
* generating the syntax validation code for an intermediate language

That'll provide a solid base for quick and easy iteration.

## Thoughts on Testing

There's the general question of how testing passes should work, as in:
1. should the pass output be compared against an expected version?
2. should the pass output be compiled down to bytecode and then run?

Right now, the second option is chosen, but I think it should be a mixture of both. Whether a
language should be compiled down into bytecode and then run should be configurable via a
command line option, so that during local development, only a pass' output is checked, whereas
in CI, it's also fully compiled and run.

Fully compiling the output of a pass during testing could work by serializing the output to
disk and then treating it as a test file for the runner of the lower-level language, repeating
the process until reaching the VM tester. This has the benefit of not having to implement the
whole pipeline into every runner, but the serialization + file IO + deserialization overhead
might be too much.

To keep the cost of changing languages at a reasonable level, I think tests should stick to
only covering each language feature in isolation, at least for now.

zerbina:
@saem: Regarding the naming, given that there can (and will) be multiple target languages, I believe having `L0` refer to the source language would be better (thought maybe a bit confusing, since lowering then corresponds to the number going up).

However, since we're developing the languages bottom-to-top, having `L0` means "target language" is easier for now, I'd say, otherwise a renaming is necessary whenever adding a new higher-level IL. For the in-development top-level candidate, we could use `Lx` (or similar) as the name.

saem:
> @saem: Regarding the naming, given that there can (and will) be multiple target languages, I believe having `L0` refer to the source language would be better (thought maybe a bit confusing, since lowering then corresponds to the number going up).
> 
> However, since we're developing the languages bottom-to-top, having `L0` means "target language" is easier for now, I'd say, otherwise a renaming is necessary whenever adding a new higher-level IL. For the in-development top-level candidate, we could use `Lx` (or similar) as the name.

I think since we're going to be stacking languages on top, and the earlier (bottom) languages are going to be around longer, it's fair to assume they'll be more stable over time. I think starting the numbering where the bottom is `L0` is likely to be the easiest (fewest renames and most stable references over time).

Although, if at some point we need a secondary numbering scheme, that does go from source -> target, then we could maybe do `S0` for source-zero, and then increment the number as it gets lower. This would be the opposite of the `Lx` scheme, but it would allow us to reason about depth from source, if and when required.

passes/trees.nim
```diff
@@ -0,0 +1,173 @@
...
+    if not isAtom(t.nodes[i].kind):
+      last += t.nodes[i].val
+    inc i
+
+  result = NodeIndex(i)
+
+proc child*(t: PackedTree, i: NodeIndex, child: Natural): NodeIndex =
+  ## Returns the index of the `child`-th subnode of `i`.
+  var i = uint32(i) + 1
+  for j in 0..<child:
```
saem, line 47:
```suggestion
  for _ in 0..<child:
```

Only because it's unused, at first I thought the loop body had a bug because it wasn't using `j`, but I glossed over the `var I` declaration above; this change might make things a bit more clear.

passes/trees.nim
```diff
@@ -0,0 +1,173 @@
...
+  ## Returns the index of each child node of the node at `at`.
+  var n = NodeIndex(ord(at) + 1)
+  for i in 0..<t.len(at):
+    yield n
+    n = t.next(n)
+
+iterator items*(t: PackedTree, at: NodeIndex; start: int; last = ^1): NodeIndex =
+  ## Returns the index of each child node of the node at `at`, only
+  ## including subnode in slice ``start..last``.
+  var n = NodeIndex(ord(at) + 1)
```
saem, line 80:
```suggestion
  assert start <= last, "cannot start after the last node"
  var n = NodeIndex(ord(at) + 1)
```

passes/trees.nim
```diff
@@ -0,0 +1,173 @@
...
+  PackedTree*[T] = object
+    ## Stores a node tree packed together in a single sequence.
+    nodes: seq[TreeNode[T]]
+    numbers: seq[uint64]
+    # TODO: use a BiTable for the numbers
+
+  NodeIndex* = distinct uint32
+
+const
+  ExternalFlag = 0x8000_0000'u32
```
saem, line 25:
```suggestion
  ExternalFlag = 0x8000_0000'u32
    ## use the most significant bit to flag whether the value is larger than an
    ## `max(int32)` and overflows into `PackedTree.numbers`.
```

saem (commented):
Partial review, notably covering `trees`, `spec`, and `lang0.md`.

tests/pass0/t03_basic_stack.test
```diff
@@ -0,0 +1,18 @@
+discard """
+  output: "(Done 4100)"
```
saem, line 2: Note to self: check why this address is stable

tests/pass0/t03_drop.test
```diff
@@ -0,0 +1,19 @@
...
+(TypeDefs
+  (Int 8)
+  (ProcTy (Void))
+)
+(GlobalDefs)
+(ProcDefs
+  (ProcDef (Type 1) (Locals)
+    (Continuations
+      (Continuation (Params) 0
+        (Drop (IntVal 8))
```
saem, line 13: Yes, but from how many meters up? ;)

tests/pass0/t03_select_bool.test
```diff
@@ -0,0 +1,22 @@
...
+"""
+(TypeDefs
+  (Int 8)
+  (ProcTy (Type 0))
+)
+(GlobalDefs)
+(ProcDefs
+  (ProcDef (Type 1) (Locals) (Continuations
+    (Continuation (Params) 0
+      (SelectBool (IntVal 1) (Continue 1) (Continue 2))
```
saem, line 12: Maybe I'm misreading/understanding the code, my impression was that it was `SelectBool <bool_expr> <goto_true> <goto_false>`, wouldn't `IntVal 1` result in a truthy value and therefore we'd go to continuation 1, which is `(IntVal 1)` and not `(IntVal 2)`?

---

N.B.: I haven't read the implementation yet, so the uncertainty around `SelectBool` below might be answered then.

saem (commented):
Partial review, went through the tests, many of the comments might be clarified upon reviewing the `pass0` module.

passes/lang0.md
```diff
@@ -0,0 +1,105 @@
...
+        |  (Sub <type_id> <value> <value>)
+        |  (Mul <type_id> <value> <value>)
+        |  (Div <type_id> <value> <value>)
+        |  (Mod <type_id> <value> <value>)
+        |  (AddChck <type_id> <value> <value> <local>)
+        |  (SubChck <type_id> <value> <value> <local>)
+        |  (Not <value>)
+        |  (Eq <type_id> <value> <value>)
+        |  (Lt <type_id> <value> <value>)
+        |  (Le <type_id> <value> <value>)
```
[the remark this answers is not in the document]
zerbina: I think having a `Gt` and `Ge` only makes sense in a higher-level language. The lowering might need temporaries (in order to keep the correct evaluation order), and not having any hidden introduction of new locals by `pass0` was an explicit goal of mine.

tests/pass0/t03_basic_local.test
```diff
@@ -0,0 +1,17 @@
...
+  (ProcTy (Type 0))
+)
+(GlobalDefs)
+(ProcDefs
+  (ProcDef (Type 1) (Locals (Type 0)) (Continuations
+    (Continuation (Params) 0
+      (Asgn (Local 0) (IntVal 1))
+      (Continue 1 (Copy (Local 0)))
+    )
+    (Continuation (Params (Type 1)))
```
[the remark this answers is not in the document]
zerbina: Yep, it has to be `(Type 0)`, good catch. `pass0` ignores the return type, so it doesn't make an effective difference (`(Type 10)` could be use an the program would compile and run successfully).

Continuation parameter types for the return continuation are a leftover from an earlier iteration, where there were no `ProcTy`pes. Higher-level languages will need the type on the return continuation, so I figured it makes sense to just keep it in lang0.

tests/pass0/t03_basic_stack.test
zerbina: All virtual address in the VM start at 4096 (accessing anything below that results in an access violation). The in-memory stack is located at the very start of the VM's memory region, and since the procedure is the very first procedure to run and allocate stack-space, its stack-pointer will be 4096. `(Addr (IntVal 4))` returns the stack-pointer + 4, hence 4100.

Relying on implementation details like this makes the test brittle, but only making sure the test compiles and runs wouldn't guarantee that stack allocation actually works. I think this tests needs to check the generated VM bytecode instead.

tests/pass0/t03_select_bool.test
zerbina: It's inverted (i.e., the first goto is for the 'false' case). I wasn't sure which way around is more intuitive, so I went with the one that maps directly to the `Branch x 0` instruction. There's no run-time cost to swapping the meaning, though.

tests/pass0/t03_drop.test
zerbina: Since `Drop` makes the value "disappear", from pretty high up, I'd say. :D

passes/trees.nim
zerbina: Yep, that's better.

passes/trees.nim
zerbina: I'm not sure. `items` works even if start > last, and being able to specify such empty ranges *could* be useful for reducing branching or arithmetic at the callsite.

passes/trees.nim
zerbina: Thank you.

tests/pass0/t03_select_bool.test
saem: As is is fine, just takes a second, maybe a comment in the grammar at most.

passes/pass0.nim
```diff
@@ -0,0 +1,718 @@
...
+    if tree.len(it) > 1: # ignore the return continuation
+      maxStack = max(tree[it, 1].imm, maxStack)
+
+  # setup the stack pointer, if one is required:
+  if maxStack > 0:
+    # XXX: instead of allocating the maximum amount of stack space up-front,
+    #      the stack could also be grown/shrunken on a per-continuation base,
+    #      which would reduce stack-space requirements in case of nested
+    #      calls
+    # XXX: maybe move stack-managment to a higher-level IL?
```
saem, line 618:
```suggestion
    # XXX: maybe move stack-management to a higher-level IL?
```

saem (approved):
LGTM

Pretty cool how this is a whole end to end language.

tests/pass0/t03_select_bool.test
zerbina: I've marked the gotos with which case the belong to in the grammar.

zerbina:
@saem: I've addressed the review comments, extended the test coverage, made some small language changes, and fixed some bugs. There's still tests missing (arithmetic and comparison operations have no coverage yet), but I think it's okay to add them in post, so that further work depending on lang0 can commence already.

All tests now also make sure the produced bytecode matches the expected one. Beyond making the workings of `pass0` easier to understand, this also ensures that the produced bytecode doesn't silently change (when making changes to `pass0`).

tests/pass0/t04_indirect_call.test
```diff
@@ -0,0 +1,28 @@
...
+  output: "(Done 100)"
+"""
+(TypeDefs
+  (Int 8)
+  (ProcTy (Type 0))
+)
+(GlobalDefs)
+(ProcDefs
+  (ProcDef (Type 1)
+    (Locals (Type 0))
```
saem, line 11: Did you mean to pass the `IntVal 100` as a parameter rather than have it hardcoded as it is now? Not sure what the `Locals` are for.

saem (approved):
minor question, non-blocking

tests/pass0/t04_indirect_call.test
zerbina: Good catch. The locals are a leftover from a time where the VM validation layer erroneously required procedures to have at least one local, and I missed removing them before committing the tests.

PR #9, opened 2024-07-25
implement a simple disassembler for the VM

zerbina:
The disassembler takes a full `VmEnv` instance and renders it into text
that the `assembler` can turn back into bytecode. It is intended mainly
for use by the test runners.

---

## Notes For Reviewers

I've made sure locally that the disassembler works / produces the correct output. It'll be integrated into the testing pipeline as part of #7 (through integration with the `pass0` test runner).


saem (approved):
Looks good, just one typo

PR #13, opened 2024-08-04
start with implementing the source language

zerbina:
## Summary

Implement a basic grammar, translation pass (source -> L1), and test
runner for the source language. The source language currently only
consists of integer and float constants.

## Details

* add a basic language grammar for the source language
* add the `source2il` pass, for translating the source language to the
  highest-level intermediate language (currently `L1`)
* add a basic test runner for testing source language expressions

### Fixes

* fix `changesets.apply` not copying the number data storage, leading
  to overflowing integer and float values pointing to nowhere


passes/source2il.nim
```diff
@@ -0,0 +1,87 @@
...
+  var typ: TypeKind
+  block:
+    # we need to know the type of the expression, so we translate the
+    # expression using a temporary builder first
+    var bu = initBuilder[NodeKind]()
+    bu.subTree Module:
+      typ = exprToIL(t, NodeIndex(0), bu)
+    # XXX: this is bad. The Changeset/Builder API needs to support
+    #      placeholders, where one can insert placeholders, with the
+    #      actual subtree being created separately
```
zerbina, line 50: The ability to create placeholders and fill them in later is one thing that's sorely missing from NimSkull's `MirBuilder`. For lowering passes, it's *usually* not a problem, but especially `mirgen` or places where production and attribute inheritance are combined into one pass (such as is the case here) suffer from the lack of such API (the double-buffering scheme used by `MirBuilder` is an attempt to mitigate this, but it has its limits).

The interface I'm aiming for is the following:
```nim
var ph: Placeholder
bu.subTree Asgn:
  bu.add Node(...)
  ph = bu.placeholder()

bu.subTree ph, Call:
  ...
```

saem: The builder would need an enum that had one extra item to represent the placeholder, but otherwise seems pretty handy. It'd be nice if we didn't need the functionality, but after trying to rethink the analysis to avoid it, it seems impossible.

saem (approved):

PR #17, opened 2024-08-17
remove low-level subroutines

zerbina:
## Summary

Remove subroutine support from the VM and L3-0. This significantly
simplifies the VM and brings the target languages closer together, thus
reducing future work and simplifying the overall compiler architecture.

## Details

* remove the subroutine-related VM instructions
* simplify the VM's exception handling; the dedicated EH code is
  obsolete and thus removed
* adjust the `err_goto` grammar and associated tests
* remove the `Subroutine` grammar from L3-0
* remove the now-unused `Subroutine` and `List` node kinds
* remove tests for subroutines

### Motivation

The motivation behind removing low-level subroutines is two-fold:
1. they only exist for the VM target - asm.js, Wasm, and C don't have
   them -, which complicates the compiler architecture
2. they have a fairly high implementation complexity in the VM

Number 1 is the main reason for the removal. Languages without low-
level subroutines need a local + `Select` dispatcher to emulate them.

This lowering would be best be implemented at a higher-level, so that
the optimization passed can take the locals' data-flow into account,
but this requires all ILs below to support two dialects: one with and 
one without subroutines.

Alternatively, the lowering could be done at a lower level, but this
would result in the same dialect problem as the other approach
(although to a lesser degree), and it'd also result in worse code
generation.

Since low-level subroutines are only an optimization, and the VM is the
only target out of the planned ones to support it, removing them is the
overall better choice, as it reduced the surface area of ILs and the
VM, as well as making lowering passes from subroutines into dispatchers
unnecessary.


zerbina:
The removal of low-level subroutines doesn't mean that subroutines become impossible to implement, just that they don't have a specialized implementation when using the VM. Whether the specialized implementation was actually more efficient than manual branching is also questionable, since it complicated exception handling in the VM significantly.

saem (approved):

PR #21, opened 2024-08-30
add a higher-level IL to use as a target for `source2il`

zerbina:
## Summary

Add the `L10` intermediate language, which is the first IL not
structured around basic blocks (i.e., continuations). The `pass10`
pass lowers it into the `L4` language.

### Motivation

Directly translating the source language into a language with explicit
basic blocks is hard. `L10` is intended to bridge the gap between the
source language and `L4`.

### Language

The name `L10` is chosen so that there's still space for more ILs. It
does away with the basic-block centered structure, instead providing
higher- level control-flow constructs such as `if`, `case`, and
`loop`.

A `try`/`except` like facility is currently missing, meaning that
exception handling (where `Raise` is used) cannot be expressed.

Some way to delimit lifetimes of locals is also missing. Locals are
currently treated as being alive for the whole duration, which can lead
to locals that have their address taken being kept alive longer than
necessary.

Apart from the new control-flow constructs, `L10` stays close to `L4`,
so that the lowering can focus on the splitting the procedure into
basic blocks.

---

## To-Do
- [x] improve the tests (mainly the documentation)
- [x] make the `t01_loop` test work
- [x] look into improving "pinned local" handling

## Notes For Reviewers
* a `try`/`except` like facility can be added later on with ease
* for a scoping / lifetime delimiting facility, I think it makes sense to wait until the source language requirements become clearer


tests/pass10/t01_loop.expected
```diff
@@ -0,0 +1,12 @@
...
+(TypeDefs
+  (ProcTy (Void)))
+(GlobalDefs)
+(ProcDefs
+  (ProcDef (Type 0)
+    (Continuations
+      (Continuation (Params) (Locals)
+        (Continue 1 (List)))
+      (Continuation (Params) (Locals)
+        (Drop (IntVal 100))
```
saem, line 11: Sorry for what is undoubtedly a silly question, but I checked out what `opcDrop` does and I can't figure out why we want a `<value>` operand for the op code, it seems unused. Is the idea that in the future it'll be possible to specify the number of items, or the address (?), we want to drop?

zerbina: `opcDrop` pops and discard the last operand from the operand stack, it's purpose is to remove operand from the stack without having to assign them to a dummy local. The `Drop` operation has the same role as `discard`, that is, it evaluates the operand but discards  the result.

`(Drop (IntVal 100))` just means "discard the integer value 100". I've used it as form of "marker", since it's a simple, short, and side-effect-free statement. The idea is to us them to make sure that statements are placed correctly in the continuations after splitting up the statement list(s).

saem (approved):
One very minor suggestion, but g2g.

PR #23, opened 2024-08-31
passtool: report unused production rules

zerbina:
## Summary

`passtool` now reports errors for unused production rules. The
motivation is preventing mistakes and to ensure that the grammars
don't accumulate garbage.

## Details

* `trim` implements the trimming, and also handles the unused
  production rule reporting
* the top-level rule must be named `top`; `lang_source.md` and
  `lang0.md` are adjusted accordingly

### DIFF
--- a/passes/lang0.md
+++ b/passes/lang0.md
@@ -100,5 +100,6 @@ top-level node, in dedicated sections, to allow for easy and fast access to
 them.
 
 ```grammar
-top ::= (Module (TypeDefs <type>*) (GlobalDefs <type_id>*) (ProcDefs <procdef>*))
+module ::= (Module (TypeDefs <type>*) (GlobalDefs <type_id>*) (ProcDefs <procdef>*))
+top ::= <module>
 ```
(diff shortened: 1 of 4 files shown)


passes/lang0.md
```diff
@@ -100,5 +100,5 @@ top-level node, in dedicated sections, to allow for easy and fast access to
 them.
 
 ```grammar
-module ::= (Module (TypeDefs <type>*) (GlobalDefs <type_id>*) (ProcDefs <procdef>*))
+top ::= (Module (TypeDefs <type>*) (GlobalDefs <type_id>*) (ProcDefs <procdef>*))
```
saem, line 103:
```suggestion
module ::= (Module (TypeDefs <type>*) (GlobalDefs <type_id>*) (ProcDefs <procdef>*))
top ::= <module>
```

Any reason not to go this route like in `lang_source`, might be slightly more self-documenting. This isn't a blocker, feel free to ignore.

saem (approved):
Good idea, since we're likely to have a lot of grammars it makes sense that people won't be intimately familiar with all aspects of them. This should avoid a lot of unnecessary detective work confirming that some rule is never used.

passes/lang0.md
zerbina: Yea, that's better.

PR #25, opened 2024-09-01
lang: implement `==`, `<=`, `<`, and `not` operators

zerbina:
* implement the basic relational operators for integers, floats, and
  booleans
* implement the logical `not` operator
* implement the literal `true` and `false` values


passes/source2il.nim
```diff
@@ -30,20 +31,117 @@ type
...
+
+    result = tkBool
+  else:
+    bu.add Node(kind: IntVal)
+    result = tkError
+
+proc notToIL(c; t; n: NodeIndex, bu): TypeKind =
+  if t.len(n) != 2:
+    bu.add Node(kind: IntVal)
+    return tkError
```
saem, line 120: Might be worth templatizing this code similar to `checkMinSons` either now or later.

saem (approved):

PR #26, opened 2024-09-05
source2il: target `L4`

zerbina:
## Summary

Target the `L4` with `source2il`, so that further development of the
source language is made easier.

## Details

* adjust the generation of procedure wrapper for expressions
* apply the necessary lowerings in the `expr` test runner
* fix a bug with `pass3`, where the `Float` type wasn't considered
  during type lowering


saem (approved):
🤯 that's insane that it's that small a change.

PR #28, opened 2024-09-06
start with an in-text language specification

zerbina:
Add an in-text specification for the source language. It currently only
covers basic terminology and the current type theory (`unit` and `void`
aren't implemented yet).

The idea is to have some written document(s) accompanying the
specification tests, to better capture intent.


spec/semantics.md
```diff
@@ -0,0 +1,33 @@
...
+A *program* consists of a single *module*. A *module* consists of zero or more
+*declarations*.
+
+A *declaration* introduces a new entity, one that can later be referred to with
+the given name.
+
+An *expression* is a term or a control-flow instruction. All expressions have a
+*type*.
+
+A *statement* is a computation without a type.
```
saem, line 14: I guess this would cover definitions like procedures, which we're not consider to be `unit`.

saem (approved):

PR #30, opened 2024-09-09
source2il: separate `Context` creation

zerbina:
## Summary

Separate `Context` creation from translation (`exprToIL`). This is a
prerequisite for incremental processing of modules.

## Details

* rename `Context` to `ModuleCtx`
* the new `open` procedure creates a new `ModuleCtx`
* the new `close` procedures finishes the `ModuleCtx` and yields the
  produced IL code
* `exprToIL` appends a proc to `Context` and only returns the type  
* `Literals` can now be appended to directly, without a `PackedTree` 
* `Context` now accumulates its own `Literals` DB, instead of re-using
  that of the input tree(s)

---

## Notes For Reviewers
* part of the work of introducing modules and procedure declarations

### DIFF
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -26,6 +26,7 @@ type
 
   Context* = object
     ## The translation/analysis context for a single module.
+    literals: Literals
     types: Builder[NodeKind]
       ## the in-progress type section
     numTypes: int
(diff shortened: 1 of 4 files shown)


passes/source2il.nim
```diff
@@ -24,10 +24,13 @@ type
     tkInt
     tkFloat
 
-  Context = object
+  Context* = object
```
saem, line 27:
```suggestion
  ModuleCtx* = object
```

You don't have to make the change, just thought it was slightly more descriptive as to what it is.

saem (approved):

PR #31, opened 2024-09-10
tester: implement `reject` feature

zerbina:
## Summary

* runners can now signal "test valid and there's an error" with exit
  code 2
* introduce the `reject` test spec key -- it defaults to 'false'

## Details

The tester now makes a distinction between:
1. valid tests that result in an error (e.g., semantic analysis for
   some piece of code failed)
2. invalid test (e.g., syntax error in test code), or runner crash

The idea with adding an extra bit of information (instead of just
re-using `output`) is to make writing simple "expect failure" tests
easier.

Aside from crashes, it's up to a runner to decide which category an
error falls into. Exit code 2 signals that there was an error that
falls into the first category.

The `reject` key specifies whether a "valid test but an error is
reported" result is expected.

### DIFF
--- a/tools/tester.nim
+++ b/tools/tester.nim
@@ -39,23 +39,28 @@ type
     ## expect from it.
     arguments: seq[string]
       ## extra arguments to pass to the runner
+    reject: bool
+      ## whether the test runner is expected to report a non-crash failure
     expected: seq[OutputSpec]
       ## the output(s) expected from the runner (if any)
     knownIssue: Option[string]
       ## whether the test is currently expected to fail due to a known issue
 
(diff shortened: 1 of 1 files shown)


zerbina:
The `reject` feature is similar to `testament`'s `action: reject`, with the main difference being that `testament` would consider the compiler crashing (instead of reporting a proper error) to satisfy the requirement.

While the feature could also be implemented by the individual runners that need it (e.g., by passing an extra argument to the runner), I think it's general enough to warrant dedicated `tester` support.

saem (approved):

PR #35, opened 2024-09-12
lang: add `Unreachable` and `void` procedures

zerbina:
## Summary

* add the `Unreachable` expression
* add the `VoidTy` type expressions
* support procedures with a `void` return type

## Details

`Unreachable` is intended to mark control-flow paths as dead. Future
progress on the language might render it obsolete again, but for now,
it fills a hole in the language and, in the absence of a `raise`-like
construct, makes `void`-return procedures (and thus `void` call
expressions) possible.

---

## Notes For Reviewers
* depends on #34

### DIFF
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -74,7 +74,7 @@ expr += (IntVal <int>)
 
 The `IntVal` expression always has type `int`, `FloatVal` always type `float`.
 
-#### `return`
+#### `Return`
 
 ```grammar
 expr += (Return res:<expr>?)
@@ -88,6 +88,23 @@ is reported if:
 
 The type of the `Return` expression is `void`. It returns control from the
 current procedure to its caller.
+
+#### `Unreachable`
+
+```grammar
+expr += (Unreachable)
+```
+
+Marks a control-flow path as unreachable. In case program execution does reach
+the `Unreachable` expression, the program immediately terminates. A compiler
+*may* report an error if it can statically detect that control-flow can reach
+an `Unreachable` expression within a procedure.
+
+> TODO: consider lifting the requirement that the program must terminate, in
+> order to allow for the compiler optimizing the affected control-flow paths
+> away
+
+The type of the `Unreachable` expression is `void`.
 
 #### Calls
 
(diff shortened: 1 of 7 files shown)


zerbina:
> Future progress on the language might render it obsolete again,

What I mean by this is that the introduction of a expression `raise` or a built-in `terminate`/`quit` procedure could allow for implementing `Unreachable` as a procedure in library code, much like it's done for the `unreachable` procedure from `vm/utils.nim`.

`Unreachable` does express something that the aforementioned implementation wouldn't, namely that some point in the program is not ever reached, so maybe it should stay, I'm not entirely sure yet.

spec/spec.md
```diff
@@ -89,6 +89,23 @@ is reported if:
...
+```
+
+Marks a control-flow path as unreachable. In case program execution does reach
+the `Unreachable` expression, the program immediately terminates. A compiler
+*may* report an error if it can statically detect that control-flow can reach
+an `Unreachable` expression within a procedure.
+
+> TODO: consider lifting the requirement that the program must terminate, in
+> order to allow for the compiler optimizing the affected control-flow paths
+> away
```
saem, line 105: Right now it states that `program execution` terminates, I'd imagine in the case of wanting to not terminate for the purposes of optimization, that would be `program analysis` of some variety that's perhaps doing some sort of pseudo evaluation?

saem (approved):
A minor clarification/confirmation, but it's not at all blocking and not all that pressing to be answered. PR looks good to me.

spec/spec.md
zerbina: > that would be program analysis of some variety that's perhaps doing some sort of pseudo evaluation?

Hm, I'm not sure I get what you mean.

In my mind, the statement "the program immediately terminates" requires that execution of program terminates -- *always*. This precludes optimizations where the compiler produces no code for control-flow paths leading up to `Unreachable`, under the assumption that `Unreachable` really does mean "never reached".

Having said that, I'm no longer sure that such an optimization is really are a good idea. Perhaps in an explicit `-d:danger`-like mode, where the is allowed to violate the specification, but not otherwise

I'll remove the TODO.

zerbina: Ah, yeah, your interpretation makes sense.

PR #36, opened 2024-09-13
lang: add tuple types and tuples

zerbina:
## Summary

* implement tuple type constructors
* implement tuple constructors
* implement general support for aggregate/complex types for `source2il`
  and the tester

### Restructuring

* since types can be more complex now, they're represented by the new
  `SemType` type
* move the type representation and corresponding query routines to the
  new `types` module

### Source2IL

* implement tuple and tuple type constructor support
* implement support for returning aggregate/complex types from
  procedures. They currently need to be returned via an out parameter
* implement expression-list lowering. They're needed for the out
  parameter handling, but the `L10` doesn't support them yet

### VM

* support setting the start of the stack with `initThread`
* implement support for aggregate return values for `vmexec.run`

### DIFF
--- a/common/vmexec.nim
+++ b/common/vmexec.nim
@@ -12,6 +12,85 @@ import
     utils
   ]
 
+proc readInt(p: HostPointer, size: range[1..8]): int64 =
+  copyMem(addr result, p, size)
+
+proc readFloat64(p: HostPointer): float64 =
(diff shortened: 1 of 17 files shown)


saem (commented):
Mini-review, nothing major, of course.

saem (approved):

PR #39, opened 2024-09-18
lang: add tuple access

zerbina:
Extend the source language with the `FieldAccess` operator, for
retrieving values from a tuple. Only static indexing with immediate
integer values is supported at the moment.

### DIFF
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -405,6 +405,37 @@ proc exprToIL(c; t: InTree, n: NodeIndex, bu, stmts): SemType =
       # it's a unit value
       bu.add Node(kind: IntVal)
       result = prim(tkUnit)
+  of SourceKind.TupleAccess:
+    let
+      (a, b) = t.pair(n)
+      tup = c.exprToIL(t, a)
+    case tup.typ.kind
+    of tkTuple:
+      let idx = t.getInt(b)
+      if idx >= 0 and idx < tup.typ.elems.len:
+        result = tup.typ.elems[idx]
+
+        case result.kind
+        of ComplexTypes:
+          bu.subTree Field:
+            bu.inline(tup, stmts)
+            bu.add Node(kind: Immediate, val: idx.uint32)
+        else:
+          # TODO: wrapping the expression in a Copy operation needs to happen
+          #       at the callsite (because only the consumer knows whether to
+          #       copy or not)
+          bu.subTree Copy:
+            bu.subTree Field:
+              bu.inline(tup, stmts)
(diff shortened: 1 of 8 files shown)


zerbina:
I'm not too happy with the name `TupleAccess`. A more general name, like `GetField` (which would also apply `object`-like types, once those are implemented), might be better.

zerbina:
The test failure is due to a bug with `vmvalidation`. I'll make a separate PR for the fix.

passes/source2il.nim
```diff
@@ -405,6 +405,37 @@ proc exprToIL(c; t: InTree, n: NodeIndex, bu, stmts): SemType =
       # it's a unit value
       bu.add Node(kind: IntVal)
       result = prim(tkUnit)
+  of SourceKind.TupleAccess:
```
saem, line 408: Would `FieldAccess` work better? It can apply to both tuples and objects.

---

#### Brain dump about compound types and fields

This is based off of NimSkull, but here are the types and how we access them and various relevant properties.

|Type|Operator|Action|Subcomponent Name|Ordered|Width|
|-----|---------|------|---------------|--------|-----|
|array|`[]`|index|element|yes|fixed|
|tuple|`[]`|index|element|yes|variable|
|tuple|`.`|access|field|yes|variable|
|object|`.`|access|field|no|variable|

I've also seen the term "subscript" used for array access, but I consider that to be somewhat off; although the advantage is that it frees up the term `index` to act as a generalization over `subscript` and `field access`. As long as we're not looking to unify access across array, object, and tuples, then I think `FieldAccess` works, especially if we're considering mapping it to the `.` operator in some top level syntax.

spec/spec.md
```diff
@@ -156,6 +157,21 @@ and so on.
...
+
+```grammar
+expr += (TupleAccess tup:<expr> index:<int_val>)
+```
+
+Retrieves the value from the `index`-th position of the tuple.
+
+Let `T` be the type of `tup`. If `T` is not a `tuple` type, an error is
+reported. If `index` is an integer value less than 0, or greater than or equal
+to the number of types in the tuple type `T`, an error is reported.
```
saem, line 170:
```suggestion
to the number of positions in the tuple type `T`, an error is reported.
```

I'm suggesting `positions` because it's possible the reader might thing `types` is a set, so `(int, int)` would have 2 `positions`, but 1 `types`.

saem (approved):
Overall the change and tests look good. I left a name suggest for `TupleAccess` and a tweak to some `spec.md` wording -- happy to discuss.

PR #43, opened 2024-09-22
source2il: remove `Asgn` lowering

zerbina:
## Summary

Simplify `source2il` by removing the in-place assignment lowering for
complex types.

## Details

Assignments for complex types are supported in the `L10`, so the
lowering is no longer necessary.

The special-casing of complex types is removed for assignments. In
addition, wrapping an lvalue expression in a `Copy` operation is now
handled at where the expression is *embedded* (via the new `genUse`)
instead of where it's *produced*.

---

## Notes For Reviewers
* made possible by #42

### DIFF
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -176,23 +176,20 @@ proc newTemp(c; typ: SemType): uint32 =
   result = c.locals.len.uint32
   c.locals.add typ
 
+proc genUse(a: NodeSeq, bu) =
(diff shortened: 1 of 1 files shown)


zerbina:
Removing the `Addr` operations from the `L10` output also improves the quality of the final bytecode, since taking the address of a local unconditionally disables the SSA form for it, which prevented some optimizations.

It's not yet possible to create code eligible for such optimization with the source language, so there's no effective difference at the moment.

saem (approved):

PR #44, opened 2024-09-24
lang: add type aliases

zerbina:
## Summary

Add the `TypeDecl` declaration to the source language, for introducing
type aliases.

## Details
 
* implement the `TypeDecl` declaration and identifiers in type
  expressions
* add the `Entity` type and adjust lookup to be able handle different
  kinds of entities (instead of just procedures)
  * instead of a case object, the separate entity kinds use separate
    types and separate sequences
  * this allows for faster and easier access and a more terse memory
    layout
* change `declToIL` to not return anything. Querying the return type
  of procedure has to be done separately now
* adjust the tester and REPL to the `declToIL` change
* fix `source2il` crashing for erroneous type expressions in return
  type positions

---

## Notes For Reviewers
* the `Entity`-related changes are also a prerequisite for local variables and parameters


zerbina:
An observation regarding the tests: the number prefix naming system for language tests doesn't seem like it'll work out. Grouping by feature seems better, and the dependencies are also not clear (e.g., what language feature does/can a `t06_x` test actually rely on?).

Of course, it could also be that I'm just holding it wrong and that a feature-centric language testing would actually be worse -- I'm not fully sure.

saem (approved):

PR #47, opened 2024-09-26
spec: prepare for variables and parameters

zerbina:
## Summary

* subdivide expressions into l-value and r-value expressions
* add definitions for values, locations, and cells
* introduce the notion of linear, affine, and normal types

## Details

Some definitions and terminology was missing to be able to describe /
specify the semantics of variables and assignments. This commit aims to
extend the specification with the necessary parts.

As a preparation for future resource management, linear and affine
types are introduced already, though they're not of practical relevance
yet.

---

## Notes For Reviewers
* I made this a separate PR from locals, so that discussion (if any) can focus on the terminology
* merging this can wait until #46 is merged 

Overall, I'm not too happy with the current specification. I think it's not formal enough, and the way it's currently structured is also off. It'll need a redo, eventually.


spec/spec.md
```diff
@@ -58,6 +66,40 @@ the same type.
...
+A *value* is something that inhabits a type. They're either constructed
+implicitly or are produced by terms. An *aggregate value* is a value
+inhabiting a composite type.
+
+*Values* are stored in *locations*. If a location stores an aggregate value,
+it consists of one or more sub-locations. A location not part of any other
+location is called a *cell*.
+
+Every implicitly or explicitly constructed value has a unique *identity*,
+meaning two separate values cannot be the *same*, but they can be *equal*.
```
zerbina, line 80: Overloading the term "value" like this (i.e., having it also refer to a run-time *instance* of a value) is a bad idea, in my opinion. A separate term would be better, though I haven't been able to come up with a satisfactory one yet.

A possible candidate could be "value instance". C uses the term ["object"](https://cppreference.com/en/c/language/object.html) for this concept.

saem (approved):

zerbina:
@saem: I've replaced the relevant mentions of `value` with `object`, which resulted in a small improvement in overall clarity of the affected sections, I believe, though it's still not ideal.

`object` needs to be replaced with something else at some point, but I agree with you that it's a serviceable term for now.

saem (approved):
Agreed, it's easier to follow after the change.

PR #48, opened 2024-09-28
pass10: fix propagation of needed locals

zerbina:
## Summary

Fix a bug with propagation of needed locals in `pass10` that resulted
in crashes for locals needed across a loop edge.

## Details

The test for detecting the loop start during backward propagation
was wrong. It triggered immediately for the basic-block ending in
the loop edge, thus disabling reiteration for the rest of the loop's
basic-blocks, which resulted in incorrect `needs` sets.

The condition is corrected and a test case for reproducing the issue is
added.


saem:
FYI: fixed a tiny typo in the PR body.

saem (approved):

PR #50, opened 2024-10-01
lang: add expression lists

saem:
## Summary

Add `Exprs` construct, which allows for expression lists, where a series of expressions can be treated as a single expression. This allows writing zero or more `unit` or `void` expressions which must be followed by a final expression of any type.

## Details

Added the spec `Exprs <exprs>+` construct support in the specification, as well as the implementation and associated tests. With the possibility of `void` expressions in non-trailing positions, `sourceToIL` translation will drop all trailing expressions after the `void` expression. This avoids `L10` errors related to unreachable code following any `void` expression (e.g.: `break`, `continue`, `return`, etc).

As part of this change the following convenience routines were added to assist in `Exprs` and future changes:
- `add` where takes a sequence of statements and adds them to a builder in succession
- `genDrop` which generates a `Drop` statement wrapper for a given expressions

N.B.: Since we neither have runtime effects (i.e.: `echo`) nor mutable local variables, some behaviour like multiple observable expressions cannot be tested as part of this change. Those tests will be added in future changes once we can sense such changes.

---

## Note for Reviewers

- [ ] ~write tests, needs effects or locals~
- [x] code changes:
   - [x] move `Exprs` decl with other non-`void` decls
   - [x] inference rules description in spec
   - [x] remove `tmp` usage
- [x] Tests to add:
   - [x] single item expression lists
   - [x] invalid expression lists with non-void, non-unit types in leading positions
   - [x] valid expression lists ending in a `void` expression
   - [x] valid expression lists ending in a non-`void` expression
   - [x] expression lists ending in a non-`void` expression, but with a `void` expression somewhere earlier


passes/spec_source.nim
```diff
@@ -20,11 +20,12 @@ type
     Params
     ProcDecl
     TypeDecl
+    Exprs
```
zerbina, line 23: Minor nit-pick: I'd move `Exprs` to after `FieldAccess`, so that it's grouped together with the other non-void expressions.

zerbina (commented):
I did a first pass and left a few comments.

Regarding the specification, it says that non-trailing expressions can be `void`, which I think makes sense for the sake of language ergonomics. Nonetheless, it's currently not specified what exactly happens in that case.

Personally, I'd prefer if such an expression list is always typed as `void`, because that side-steps a good amount of trouble and implementation complexity.

spec/spec.md
```diff
@@ -179,6 +179,17 @@ to the number of positions in the tuple type `T`, an error is reported.
...
+#### Expression Lists
+
+```grammar
+expr += (Exprs <expr>+)
+```
+
+A non-empty list of expressions, where the tail expression may be any type and
+preceeding ones must be `unit` or `void`. An error is reported if:
+* the expression list is empty
+* a non-tail expression is not of `unit` or `void` type
```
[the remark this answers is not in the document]
saem: I'll add in the inference behaviour, specifically it takes the last expression's type, unless there is a preceding `void`, in which case it's `void`.

passes/source2il.nim
```diff
@@ -571,6 +573,31 @@ proc exprToIL(c; t: InTree, n: NodeIndex, bu, stmts): SemType =
...
+    var eb = bu
+    stmts.addStmt Stmts:
+      for i, si in t.pairs(n):
+        let e = c.exprToIL(t, si)
+        for s in e.stmts:
+          bu.add s
+        case e.typ.kind
+        of tkUnit:
+          bu.subTree Drop:
+            bu.add e.expr
```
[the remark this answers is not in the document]
saem: Hmm, do you think it's worth adding a `drop` routine, that does the `subtree Drop: genUse(...)`?

passes/source2il.nim
```diff
@@ -571,6 +573,31 @@ proc exprToIL(c; t: InTree, n: NodeIndex, bu, stmts): SemType =
...
+        else:
+          if i != last:
+            c.error("non-trailing expressions must be unit or void, got: $1" %
+                      [$e.typ.kind])
+        if i == last:
+          if e.typ.kind != tkVoid:
+            let tmp = Node(kind: Local, val: c.newTemp(e.typ))
+            c.genAsgn(tmp, e.expr, e.typ, bu)
+            eb.add tmp
```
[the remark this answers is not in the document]
saem: Oh of course, it's already lowered via `exprToIL`, so it'll be understandable to `L10`, as in it's already had any sort of local assignment done if it needed it in the first place.

spec/spec.md
```diff
@@ -179,6 +179,19 @@ to the number of positions in the tuple type `T`, an error is reported.
...
+```grammar
+expr += (Exprs <expr>+)
+```
+
+A non-empty list of expressions, where the tail expression may be any type and
+preceding ones must be `unit` or `void`. An error is reported if a a non-tail
+expression is not of `unit` or `void` type.
+
+The type of the expression list is inferred as `void` if any non-trailing
+expression is `void`, otherwise the type is that of the trailing expression.
```
saem, line 193: A realization on my part/small tidbit:

In Haskell an expression/routine type is not a Set, but instead a Hask, which is a Set + a bottom indicating it might never return. We're headed down a similar road, where `void` sub-types everything means the same. Also, in case anyone is wondering I did think about the implication of dropping this, but there are very few instances where you'd want to ensure an expression is guaranteed to fully evaluate. Even then many of those can be handled through other means, like ensuring all errors are handled (read: no cause for abnormal termination).

passes/builders.nim
```diff
@@ -48,6 +48,15 @@ func add*[T](bu: var Builder[T], nodes: openArray[TreeNode[T]]) =
...
   bu.buf.add nodes
 
+func add*[T](bu: var Builder[T], trees: openArray[seq[TreeNode[T]]]) =
+  ## Appends all `trees` to the current sub-tree. The trees must each either
+  ## represent a single atomic node, or a complete subtree.
+  assert trees.len > 0
+  if bu.parent != -1:
+    inc bu.buf[bu.parent].val
+  for t in trees.items:
+    bu.buf.add t
```
saem, line 58: Not sure if we should add this overload, but it came up here and a bunch in the `If` branch. I think this is going to be a relatively common operation, but there is some minor merit to seeing the explicit `for ...`. This isn't a hard requirement for me.

passes/source2il.nim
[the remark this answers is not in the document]
zerbina: Yep, I think it's worth adding -- this code fragment will probably come up again a few times.

passes/builders.nim
zerbina: I think it makes sense to have, but given that it's mostly useful to `source2il`, I'd move it there. For ergonomic reasons, I'd also drop the assertion (so that passing empty statements lists to `add` works).

saem: Cool, I moved it over and dropped the assertion plus some other inner builder access.

zerbina:
I've gathered a list of things that need testing, but where the tests don't require effect-having expressions:
* single item expression lists
* invalid expression lists with non-void, non-unit types in leading positions
* valid expression lists ending in a `void` expression
* valid expression lists ending in a non-`void` expression
* expression lists ending in a non-`void` expression, but with a `void` expression somewhere earlier

saem:
Right now the following:

```
(Exprs (Unreachable) (TupleCons) (FloatVal 1.0))
```

Fails because it results in the following `L10`, which has unreachable code errors:

```
(Unreachable)
(Drop (IntVal 0))
(FloatVal 1.0)
```

I'm conflicted as to what to do:

1. remove the restriction
2. remove the flag `Stmt` restriction
3. have `Exprs` processing culling everything after `Unreachable` and consider that desirable behaviour
4. accept this as a failing test and update the spec, if needed

zerbina:
The assertion message is maybe a bit unclear in this regard, but it triggers for all dead code situations, not just `Unreachable`. For example, the following code also trigger the assertion:
```nim
(ProcDecl (Ident "a") (UnitTy) (Params)
  (Stmts
    (Return)
    (TupleCons)))
```

Now, the assertion could be removed without causing any issues, as `pass10` would just discard the dead code following the `Return`.

In general, I prefer aborting/crashing over silent error correction in case of easily detectable issues with the input, though I'm not sure whether dead code should really be considered an error in the `L10` language (it's currently not described as such).

> I'm conflicted as to what to do, some options I can think of:
> 1. remove the `Unreachable` restriction
> 2. remove the flag `Stmt` restriction
> 3. have `Exprs` processing culling everything after `Unreachable` and consider that desirable behaviour
>  4. accept this as a failing test and update the spec, if needed

My personal preference is towards number 3, though I think 4 would be fine too. If unreachable code being a hard error turns out to be too much of an ergonomic problem, then we can change it later on and allow `void` expression in non-trailing positions again.

saem:
> The assertion message is maybe a bit unclear in this regard, but it triggers for all dead code situations, not just `Unreachable`. For example, the following code also trigger the assertion:
> 
> ```nim
> (ProcDecl (Ident "a") (UnitTy) (Params)
>   (Stmts
>     (Return)
>     (TupleCons)))
> ```

Ah, good point, I took the message too literally, a break, continue, raise, non-terminal call, etc would likely all result in the same.

> Now, the assertion could be removed without causing any issues, as `pass10` would just discard the dead code following the `Return`.
> 
> In general, I prefer aborting/crashing over silent error correction in case of easily detectable issues with the input, though I'm not sure whether dead code should really be considered an error in the `L10` language (it's currently not described as such).

That's why I was conflicted, I thought the error/failure had benefit.

> > I'm conflicted as to what to do, some options I can think of:
> > 
> > 1. remove the `Unreachable` restriction
> > 2. remove the flag `Stmt` restriction
> > 3. have `Exprs` processing culling everything after `Unreachable` and consider that desirable behaviour
> > 4. accept this as a failing test and update the spec, if needed
> 
> My personal preference is towards number 3, though I think 4 would be fine too. If unreachable code being a hard error turns out to be too much of an ergonomic problem, then we can change it later on and allow `void` expression in non-trailing positions again.

Looking at it today with fresher eyes, I think 3 makes more sense, too. Since the higher level layer knows _why_ it's eliminating the code (read: doing it with good reason) I think it's safe and `L10` can catch all the issues where a mistake have been made.

saem:
I didn't update the spec with non-trailing `void` expression post-expression elimination because it didn't seem like a particularly observable effect, but I could go either way.

zerbina:
> I didn't update the spec with non-trailing `void` expression post-expression elimination because it didn't seem like a particularly observable effect, but I could go either way.

I don't think it's necessary now. Maybe it'll be in the future, for things like exception tracking (if that feature makes the cut), but that's still a long time away.

zerbina (commented):
I've found two remaining issues:
1. `void` expressions in trailing positions are disallowed, even though the specification says otherwise (I've suggested the changes for fixing this)
2. expressions following a `void` expression in a list are not checked for errors

For the second problem, there are two possible solutions I can think of:
1. add a `keep` (or similar) boolean flag that defaults to true and guards every `addStmt` or `bu.add` call. On encountering a `void` expression, `keep` is set to false
2. use two loops: one that processes all expressions up-to and including the first void expression and one that processes the rest, by calling `exprToIL` and making sure the type is correct and then simply discarding the generated code

I'm fine with both, though with maybe a slight preference towards the second one.

saem:
No idea, how I convinced myself that reporting an error for a trailing `void` made sense, good catch.

For the skipped sem check of post-void expressions, I was about to use a loop, but the template approach seems to read pretty cleanly. I also renamed the tests to drop `_error` from the file name because that's reserved for testing expected errors, instead of expected output that happens to be a runtime error.

passes/source2il.nim
```diff
@@ -621,6 +635,37 @@ proc exprToIL(c; t: InTree, n: NodeIndex, bu, stmts): SemType =
...
+    for i, si in t.pairs(n):
+      let e = c.exprToIL(t, si)
+      voidGuard:
+        stmts.add e.stmts
+      if i == last:
+        voidGuard:
+          result = e.typ
+        case e.typ.kind
+        of tkVoid: discard "okay, nothing to do"
+        else:      voidGuard: bu.add e.expr
```
zerbina, line 653:
```suggestion
          case e.typ.kind
          of tkVoid: discard "okay, nothing to do"
          else:      bu.add e.expr
```

zerbina (approved):
Nicely done, good work! One very small suggestion, but this PR is good to be merged even without it.

PR #53, opened 2024-10-03
source2il: disallow using built-in name for declaration

zerbina:
## Summary

* an error is now reported when using the name of a built-in for a
  procedure or type declaration
* error messages for using entities in an incorrect context are
  improved
* `false` and `true` are now treated as built-ins
* the internal implementation of lookup is improved

## Details

The specification already said that built-in names were disallowed, but
`source2il` didn't adhere to it.

* add the `lookup` procedure, implementing lookup in a single place
* replace all manual `scope` lookup with `lookup`
* add a table that keeps track of all built-in entities
* improve error messages for unexpected entity kinds
* list `true` and `false` as names automatically part of every scope

Instead of adding the built-ins to `scope`, the constant `BuiltIns`
table is queried as a fallback when lookup in `scope` failed, which is
a tiny bit more memory efficient.

---

## Notes For Reviewers
* split-out from the work on assignments and locals

### DIFF
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -21,6 +21,9 @@ type
   NodeSeq    = seq[Node]
 
   EntityKind = enum
+    ekNone        ## non existent
+    ekBuiltin     ## some built-in special entity
+    ekBuiltinProc ## some built-in special
     ekProc
     ekType
 
@@ -67,6 +70,18 @@ type
     expr: NodeSeq
     typ: SemType
 
+const
+  BuiltIns = {
+    "+": ekBuiltinProc,
+    "-": ekBuiltinProc,
+    "==": ekBuiltinProc,
+    "<=": ekBuiltinProc,
+    "<": ekBuiltinProc,
+    "not": ekBuiltinProc,
+    "true": ekBuiltin,
+    "false": ekBuiltin
+  }.toTable
+
 using
   c: var ModuleCtx
   t: InTree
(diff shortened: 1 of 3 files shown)


passes/source2il.nim
```diff
@@ -21,6 +21,9 @@ type
   NodeSeq    = seq[Node]
 
   EntityKind = enum
+    ekNone        ## non existent
```
zerbina, line 24:
```suggestion
    ekNone        ## signals "non-existent"
```

passes/source2il.nim
```diff
@@ -77,11 +92,31 @@ proc error(c; message: sink string) =
   ## Sends the error diagnostic `message` to the reporter.
   c.reporter.error(message)
 
+func lookup(c: ModuleCtx, name: string): Entity =
+  ## Implements the lookup action described in the specification.
+  result = c.scope.getOrDefault(name, Entity(kind: ekNone))
+  # built-ins are not added to any scope, and can thus be used to detect
+  # a lookup failure here
```
zerbina, line 99:
```suggestion
```

passes/source2il.nim
```diff
@@ -21,6 +21,9 @@ type
   NodeSeq    = seq[Node]
 
   EntityKind = enum
+    ekNone        ## signals "non-existent"
+    ekBuiltin     ## some built-in special entity
+    ekBuiltinProc ## some built-in special
```
saem, line 26:
```suggestion
    ekBuiltinProc ## some built-in special procedure
```

passes/source2il.nim
```diff
@@ -21,6 +21,9 @@ type
   NodeSeq    = seq[Node]
 
   EntityKind = enum
+    ekNone        ## signals "non-existent"
+    ekBuiltin     ## some built-in special entity
```
saem, line 25:
```suggestion
    ekBuiltinVal     ## some built-in special value
```

I think it might be better to be specific and then have a catch-all in the future if we need it.

saem (approved):
Minor suggestion for renaming and some doc comments, otherwise looks good to go.

passes/source2il.nim
zerbina: Thanks, that's a better approach.

PR #55, opened 2024-10-07
lang: add locals and assignments

zerbina:
## Summary

* implement declarations for local variables (`Decl`)
* implement assignments (`Asgn`)

### Language

* scopes can now be nested
* declarations for locals are introduced
* assignments are introduced

The specification for assignments already covers the copy/move/
destruction behaviour, although there's current no way to test it.

#### Current Limitations

For the initial version, the additions are kept as simple as possible.
Notably:
* all locals are mutable (there's no way to declare them as being
  immutable)
* a local declaration *requires* an initial value for the local
* the type cannot be specified explicitly (it's always inferred)
* an assignment with an l-value RHS is always a copy assignment
* names cannot be shadowed (i.e., a local cannot use the same name as
  an already declared procedure)

The plan is to lift/remove all of the above limitations/restrictions/
shortcomings at a later date.

### Source2IL

* scope management is changed to accommodate nesting
* `exprToIL` now returns the `ExprType` type, which carries additional
  expression attributes. At the moment, this is only the `Lvalue`
  attribute
* the convenience `+` operator is added, for constructing `ExprType`
  values
* all result assignments in `exprToIL` are updated
* the convenience procedure `scopedExprToIL` is added, for analyzing an
  expression within a new nested scope

Tests are added to ensure the declarations and assignments work as
specified.

---

## To-Do
* [x] ~look into removing the "local declaration can be used as expression" rule~
* [x] merge main into the PR once #46 and #47 are merged
* [x] add tests:
  * [x] declaration with void initializer expression
  * [x] declaration with preexisting name
  * [x] declaration with self-reference
  * [x] assignment with l- and r-value lhs
  * [x] assignment with l- and r-value rhs
  * [x] assignment with incompatible type
  * [x] assignment with different but compatible type
  * [x] evaluation order in assignments
  * [x] evaluation order in expression lists


zerbina:
One thing that's also not present, when compared to NimSkull, is the style-insensitivity for identifiers. Personally, I like NimSkull's style-insensitivity rules, and they're something I would also add to Phy, but it's not something that should happen as part of this PR.

zerbina:
In order to not treat declarations as expression, the language grammar grammar needs to be extended with *groups*, so that the `Exprs` grammar can be changed to:
```
(Exprs (<local_decl>* <expr>)+)
```
`*` matches greedily, which is why `(Exprs <expr_or_decl>* <expr>)` cannot work.

I'll make a separate PR for this extension.

saem (commented):
Partial review: I've read through the logic, I left a few minor comments. I'm going to work my way through the tests next and revisit the logic one last time.

passes/source2il.nim
```diff
@@ -632,18 +716,32 @@ proc exprToIL(c; t: InTree, n: NodeIndex, bu, stmts): SemType =
...
+      # emit the destination expression in-place
+      let dst = c.exprToIL(t, a, bu, stmts)
+      if Lvalue notin dst.attribs:
+        c.error("LHS expression must be an l-value expression")
+
+      let src = c.fitExpr(c.exprToIL(t, b), dst.typ)
+      stmts.add src.stmts
+      genUse(src.expr, bu)
+
+    bu.add Node(kind: IntVal)
```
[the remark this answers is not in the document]
zerbina: Ah, yea, I had thought of that but then never got around to actually do it. Thank you for the reminder! Using a named constant makes the intention much clearer to the reader, and it also becomes easier to adjust the `unit` handling once the target language gets the `unit` type.

tests/expr/t10_asgn_lhs_must_be_lvalue_4.test
```diff
@@ -0,0 +1,5 @@
+discard """
+  description: "A literal integer is not an l-value expression"
+  reject: true
```
saem, line 3: Same `_error` suffix in test file name

tests/expr/t10_asgn_lhs_must_be_lvalue_5.test
```diff
@@ -0,0 +1,7 @@
+discard """
+  description: "A tuple constructor is not an l-value expression"
+  reject: true
```
saem, line 3: Same `_error` suffix in test file name

tests/expr/t10_asgn_lhs_must_be_lvalue_6.test
```diff
@@ -0,0 +1,7 @@
+discard """
+  description: "A projection from an r-value is not an l-value"
+  reject: true
```
saem, line 3: Same `_error` suffix in test file name

tests/expr/t10_asgn_lhs_must_be_lvalue_7.test
```diff
@@ -0,0 +1,7 @@
+discard """
+  description: "A call expression is not an l-value expression"
+  reject: true
```
saem, line 3: Same `_error` suffix in test file name

tests/expr/t11_scopes_if_2.test
```diff
@@ -0,0 +1,12 @@
+discard """
+  description: "
+    Entities declared within an `If` branch are not visible outside it
+  "
+  reject: true
```
saem, line 5: Same `_error` suffix in test file name

tests/expr/t11_scopes_local_cannot_shadow_1.test
```diff
@@ -0,0 +1,8 @@
+discard """
+  description: "Both If branches are part of their own scope"
+  reject: true
```
saem, line 3: Same `_error` suffix in test file name

tests/expr/t11_scopes_local_cannot_shadow_2.test
```diff
@@ -0,0 +1,9 @@
+discard """
+  description: "Both If branches are part of their own scope"
+  reject: true
```
saem, line 3: Same `_error` suffix in test file name

saem (approved):
Looks good, I left a soft suggest (feel free to ignore) with have `_error` be a part of test files that are expected to be rejected. Just so we can easily find by file name all tests meant to be fail.

tests/expr/t09_decl_no_self_visibility.test
```diff
@@ -0,0 +1,9 @@
+discard """
+  description: "
+    The declared local is not visible to the initializer expression
+  "
+  reject: true
```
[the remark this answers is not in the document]
zerbina: Yep, will do.

PR #59, opened 2024-10-14
pass4: fix `Rename` edge conflict resolution

zerbina:
## Summary

Fix the graph coloring failing for `Rename` edges in more complex
cases, which led to internal errors or incorrect code generation.

## Details

The two passes (one forward and one backward propagation pass) used to
color the graph are not enough to ensure that all nodes in a `Rename`
subgraph (i.e., all nodes connected to each other via `Rename` edges)
share the same color.

Consider the following graph: `A -> B; C -> B; C -> D`. After the
forward propagation pass, the graph is colored like so:
`A(1) -> B(1); C(2) -> B(1); C(2) -> D(2)`, and after the subsequent
backward pass, the result is `A(1) -> B(1); C(1) -> B(1), C(1) -> D(2)`,
which is incorrect (i.e., `D` should have the color 1 too).

### Solution

To handle arbitrarily complex graphs properly, but without resorting to
a fixpoint algorithm, the following conflict resolution is now used:
1. colors are added to buckets in a way such that all colors connected
   via rename edges are part of the same bucket
2. each bucket is assigned a color
3. all nodes whose color is in a recolor bucket are changed to the
   color of the bucket

The actual implementation is optimized such that it only requires one
`seq` and one `Table`.

---

## Notes For Reviewers
* discovered while testing the pass with some real-world code


passes/pass4.nim
```diff
@@ -213,11 +218,22 @@ proc colorGraph(gr: var Graph) =
...
           else:
             gr.nodes[e.dst].color = color
+        elif e.noCopy and gr.nodes[e.dst].color != color:
+          # problem: the source and destination should have the same color,
+          # they but don't. Register an ordered conflict pair
+          var
+            a = color
+            b = gr.nodes[e.dst].color
+          if b < a:
+            swap(a, b)
```
saem, line 228:
```suggestion
          let
            a = min(color, gr.nodes[e.dst].color)
            b = max(color, gr.nodes[e.dst].color)
```

Soft suggestion: it drops the `var` and I think is roughly as readable; also it's consistent with the `min` that comes later

passes/pass4.nim
```diff
@@ -226,21 +242,48 @@ proc colorGraph(gr: var Graph) =
...
+        a[] = m
+      if m < b:
+        map[it.b] = m
+
+    # now do the actual recoloring:
+    for i, c in gr.conts.pairs:
+      for n in gr.nodes.toOpenArray(c.nodes.a, c.nodes.b).mitems:
+        if n.keep and n.color in map:
+          let newColor = map[n.color]
+          if newColor != n.color:
```
saem, line 271: If they are the same then setting it would be a no-op, why not always set it and skip the `if`?

saem (approved):
I did a review, I should probably read through it again because I'm not sure I did a good job tracing all the logic, but I've already taken so long and the added tests add a lot of confidence.

zerbina:
> I did a review, I should probably read through it again because I'm not sure I did a good job tracing all the logic, but I've already taken so long and the added tests add a lot of confidence.

I think there's some inherent complexity to the graph coloring, and it's also not the most simple thing to visualize. Consider this graph, which corresponds to the `Graph` that `buildGraph` produces for the `three_clusters` test:
graph

The `BBx` boxes represent the basic blocks / continuations (`Graph.conts`), the nodes represent the `Params`/`Locals` on the continuations (`Graph.nodes`), and the edges represent the `Move`/`Rename`s (`Graph.edges`).

The general idea with the graph coloring is to give the same name to more than one continuation parameter / local, in order to reduce copies and to enforce the `Rename` guarantee.

### Algorithm

The first step is the forward propagation, which means iterating over the blocks from low to high, propagating the color of each node in the block along the outgoing edges (if the target node doesn't have a color already). Source nodes without a color are assigned a new unique color.

The result: 
graph_fwd

Edges where propagation failed are marked red. If the edges would come from a `Move`, there would be no problem: they're either resolved by the following backwards pass or they simply result in copies. `Rename` edges (i.e., edges where `Edge.noCopy == true`) *require* that no copy takes place, however.

The old algorithm only did a backwards propagation pass, which resulted in:
graph_bw

(But only when disabling the assertions). Since all edges are `Rename` edges in this case, this is wrong.

### New Algorithm

Before doing the backwards propagation, the new algorithm gathers all coloring conflicts as a sorted list of ordered pairs (without de-duplication), which in this case produces:

```
(1, 2) (2, 4) (2, 6) (5, 6)
```

To resolve the conflicts, all colors involved in conflicts (i.e., assigned to nodes that are connected with red edges) must be changed such that for every pair in the list, the two colors map to the same new color. If that isn't possible, the requested constraints cannot be satisfied.

passes/pass4.nim
zerbina: It's a premature optimization to prevent memory writes when there's no changes, though I don't think it's all that beneficial, nor have I done any measurements for this code. Thank you for the question -- I'll change it.

zerbina: Oh, sorry, I didn't re-read the code before commenting. The guard is necessary to prevent the assertion (which needs to be replaced with proper error reporting through `ReportContext` at some point) from triggering when it shouldn't.

passes/pass4.nim
zerbina: Good call on using a more functional style. Instead of using `min`/`max`, I've added a dedicated procedure for constructing the sorted pair, which I think conveys the intent a bit better.

passes/pass4.nim
saem: > Oh, sorry, I didn't re-read the code before commenting. The guard is necessary to prevent the assertion (which needs to be replaced with proper error reporting through `ReportContext` at some point) from triggering when it shouldn't.

I was thinking that the guard could be moved into the assertion itself and then it could always assign thereafter.

PR #60, opened 2024-10-16
lang: add user-defined procedure parameters

zerbina:
## Summary

Add support for user-defined immutable parameters (`ParamDecl`) to the
source language. Except for `void`, all types are allowed for
parameters, including `unit`.

## Details

L-value expressions are currently always copied from when passed to a
parameter. The plan is to eventually use copies only when borrowing is
not possible, but the borrow checker necessary for this is not
implemented yet. All temporaries spawned for parameter passing in a
call expression are destroyed as soon as the procedure exits.

The parameters being immutable in the called procedure allows for
safely using pass-by-reference.

Since the L30 language doesn't support aggregate parameters,
`source2il` always passes aggregate parameters by address, regardless
of size. Mutability of expressions is tracked via the new `Mutable`
expression flag.

---

## To-Do
* [x] finish the specification
* [x] implement the feature
* [x] add tests


zerbina:
The idea with making parameters immutable handles without a callee-owned location is that it allows moving all parameter binding to the callsite, which allows the implementation to turn the temporaries at the callsite into borrowing locals (according to the explicitly stated rules of the not-yet-implemented borrow checker), while still allowing fallback to a full copy.

This is essentially a rethought version of how parameters work in Nim/NimSkull, with the important difference that it is enforced that there's no conflicting mutable handle.

The (not yet specified) `Move` and `Borrow` operators can then be used to make sure that a move or borrow really takes place, with the program being ill-formed (read, an error is reported) if not. If errors are preferred over implicit copies, the rules can be changed such that implicit copies result in a static error unless suppressed with the `Copy` operator.

My aim here is for language ergonomics that can be summarized as follows:
```
In most cases, the language automatically does the "right thing", with in-language tools
being provided for ensuring that an expected action takes place.
```

### ABI

With the current design, parameters can use the same ABI as parameters in NimSkull. That is, parameters above a certain size threshold are internally passed by reference.

### Implicit moves

By way of being defined in terms of `Asgn`, parameter binding currently allows for implicit moves, but I think this should change -- there's no performance benefit to moving into the temporary (the callsite still needs to destroy the temporary after the call) and reasoning about when an object is destroyed becomes harder.

### `var` and `sink` parameters

Both are things that I think are missing (especially `var` parameters), or at least some equivalent thereof. Some `sink` like facility could be implemented right away (though for now without support for implicit moves), but for `var` parameters, I think that they should be designed together with view types.

zerbina:
In the abstract, I think the following are needed:
1. non-owning immutable parameter (the default). Can result in either a copy or a borrow at the callsite (or an error, if no copy can be made of the argument)
2. non-owning mutable borrow parameter. An error is reported at the callsite if no mutable borrow is possible
3. non-owning immutable borrow parameter. An error is reported at the callsite if no immutable borrow is possible
4. owning immutable (or mutable) parameter. The argument is automatically moved into the parameter, if possible (otherwise an error is reported)

Again, this is similar to what NimSkull provides, but with some important differences:
1. in NimSkull, a non-owning immutable parameter *never* results in a copy at the callsite
2. `var` is the closest to a non-owning mutable borrow parameter, but no static checks take place for `var` parameters in NimSkull
3. there are no explicit non-owning immutable borrow parameter in NimSkull

A `sink` parameter is equivalent to an owning mutable parameter.

While it'll makes parameters fairly limited in the beginning, I'm only going to implement non-owning immutable parameters as part of this PR -- the other parameter types can be discussed/implemented later (mutable parameters need to wait until a borrow checker is implemented, anyway).

passes/source2il.nim
```diff
@@ -1031,23 +1077,47 @@ proc declToIL*(c; t; n: NodeIndex) =
...
+    # add the parameters to the scope:
+    for i, it in t.pairs(t.child(n, 2)):
+      let name = t.getString(t.child(it, 0))
+      if c.lookup(name).kind != ekNone:
+        c.error("redeclaration of '" & name & "'")
+
+      # add the local and register the entity regadless of whether there was
+      # an error
+      if procTy.elems[i + 1].kind in AggregateTypes:
+        c.locals.add prim(tkInt) # the parameter is of pointer type internally
```
saem, line 1105: might be worth lifting this into a helper/const like creating a unit value, like `pointerParam`

saem (approved):
two soft implementation suggestions, but feel free to disregard them if you don't think they're worth it.

passes/source2il.nim
zerbina: That sounds like a good idea, at least until proper pointer types are added.

passes/source2il.nim
```diff
@@ -71,10 +72,14 @@ type
       ## the index of the out parameter, if one is required
     locals: seq[SemType]
       ## all locals part of the procedure
+    params: seq[tuple[typ: SemType, local: uint32]]
+      ## the parameters for the procedure
 
   ExprFlag {.pure.} = enum
     Lvalue ## the expression is an lvalue. The flags absence implies
            ## that the expression is an rvalue or void expression
+    Mutable## the expression refers to a mutable lvalue. Only makes sense
```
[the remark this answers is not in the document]
zerbina: I have a slight preference towards not having overlapping meaning for enum flags, so I'd personally keep both separate, at least for now.

PR #61, opened 2024-10-20
add the `phy` CLI application

zerbina:
## Summary

Bundle the passes, `source2il`, the assembler, and the VM into a single
CLI application. The idea with it is to help with development, by
providing a convenient way to invoke the passes/VM/etc. and inspect the
(intermediate) output.

## Details

The `phy` program provides a small command line interface for
specifying the source and target language are, what IRs to print, and
what to do (only compile or also evaluate).

Much of the implementation is copied from the various test runner:
* `source2il` invocation and comes from the `expr` test runner
* rendering VM execution results without high-level type information
  is moved over from the `pass0` test runner (into `vmexec.nim`)

The program also provides the `--runner` option, which makes it
behave in a way such that it's suitable as being used as a test runner,
by lifting some input restrictions and altering the output formatting,
so that it can maybe replace the currently used custom test runners at
some point.

### DIFF
--- a/phy/phy.nim
+++ b/phy/phy.nim
@@ -142,7 +142,7 @@ proc print(env: VmEnv) =
     stdout.write(disassemble(env))
 
 proc sourceToIL(text: string): (PackedTree[spec.NodeKind], SemType) =
-  ## Given a S-expression representation of the source language (`text`),
+  ## Given an S-expression representation of the source language (`text`),
   ## analyzes it and translates it to the highest-level IL. Also returns the
   ## return of the procedure to executre, or ``tkError`` if there is no
   ## procedure to run.
--- a/phy/phy.nim
+++ b/phy/phy.nim
@@ -1,0 +1,345 @@
+## Implements a CLI for running the various parts that make up Phy. This
+## includes the assembler, the various passes, `source2IL <source2il.html>`_,
+## and the VM.
+
+import
+  std/[
+    os,
+    parseopt,
+    streams,
+    strutils
+  ],
+  experimental/[
+    sexp
+  ],
+  common/[
+    vmexec
+  ],
+  generated/[
+    lang0_checks,
+    lang1_checks,
+    lang3_checks,
+    lang4_checks,
+    lang10_checks,
+    source_checks
+  ],
+  passes/[
+    changesets,
+    debugutils,
+    source2il,
+    pass0,
+    pass1,
+    pass3,
+    pass4,
+    pass10,
+    trees
+  ],
+  phy/[
+    default_reporting,
+    types
+  ],
+  vm/[
+    assembler,
+    disasm,
+    utils,
+    vmenv,
+    vmvalidation
+  ]
+
+# cannot import normally, as some type names would conflict
+import passes/spec except NodeKind
+import passes/spec_source except NodeKind
+
+type
+  Language = enum
+    langBytecode = "vm"
+    lang0 = "L0"
+    lang1 = "L1"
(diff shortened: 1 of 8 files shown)


zerbina:
Replacing the current runners with `phy --tester` would reduce code duplication (the `pass10` - `pass1` test runners are slightly adjusted copies of each other), though it also has the downside of introducing some per-test overhead, which could be a problem in practice. It's worth trying out, I think.

phy/phy.nim
```diff
@@ -0,0 +1,346 @@
...
+    stdout.writeLine(pretty(tree, tree.child(0)))
+    stdout.writeLine(pretty(tree, tree.child(1)))
+    stdout.writeLine(pretty(tree, tree.child(2)))
+
+proc print(env: VmEnv) =
+  genericPrint(langBytecode):
+    stdout.write(disassemble(env))
+
+proc sourceToIL(text: string): (PackedTree[spec.NodeKind], SemType) =
+  ## Given a S-expression representation of the source language (`text`),
```
saem, line 145:
```suggestion
  ## Given an S-expression representation of the source language (`text`),
```

phy/phy.nim
```diff
@@ -0,0 +1,346 @@
...
+        error "there's nothing to run"
+
+      if source == langSource:
+        # we have type high-level type information
+        stdout.write run(env, env.procs.high.ProcIndex, typ)
+      else:
+        # we don't have high-level type information
+        stdout.write run(env, env.procs.high.ProcIndex)
+
+main(getExecArgs())
```
saem, line 346:
```suggestion
when isMainModule:
  main(getExecArgs())
```

This would allow importing the module if one wanted. In particular, I think it would mean we could embed the functionality of this module in the runners and incur less overhead?

phy/phy.nim
```diff
@@ -0,0 +1,346 @@
...
+  of None:
+    echo HelpText
+    quit(1)
+  of Eval:
+    if target != langBytecode:
+      error "'e' requires the target language to be '" & $langBytecode & "'"
+  else:
+    discard "nothing special to do"
+
+  if ord(target) > ord(source):
```
saem, line 261: If I'm reading this correctly, everything above this line is CLI interfacing, what follows is basically all "driver" logic that runs the requested file/test. So if it makes sense to allow reusing the logic in this module then this line and below need to be lifted into its own routine.

saem (approved):
That's pretty handy, thanks.

I left a few minor comments, they're probably better off only implemented if the idea is to trim down the runner files in each passes' tests.

zerbina:
@saem: I did consider embedding (i.e., importing the module) `phy` into the runners, but there's a problem with this approach: for every runner, all passes and their satellite modules would be recompiled, instead of only the parts needed by the runners (which is how it works at the moment).

What I have in mind is to add a feature to the tester where, in addition to the current dedicated programs, a directory can specify a generic command for running the tests. For example, the `expr` tests would use the following command `bin/phy --source:source --target:vm --runner r $file`. This approach has the benefit that the `phy` executable can be compiled once and then be reused for all pass tests.

Since creating the PR, I've also measured how long the command line processing takes for `phy --source:L10 --target:vm --runner c file` on my machine (where `phy` was compiled in debug mode with `--opt:speed`), and it's ~0.015 milliseconds. For comparison, compiling `phy.nim` without the command line processing and with an empty compilation cache takes ~14.5 seconds. I think it's quite safe to say that we won't reach the amount of tests per runner for embedding the `phy` module into runners to ever amortize.

PR #63, opened 2024-10-24
tests: replace some custom runners with `phy`

zerbina:
## Summary

* support using external runner programs with the `tester`
* replace the runners for the `passX` and `expr` tests with using `phy`

## Details

If a test directory contains a `runner.txt` instead of a `runner.nim`
file,  command provided in the text file is used for running tests
from that directory. The placeholder `${file}` and `${args}` are used
for specifying where the test file and test-provided arguments are to
be inserted.

To support external runners, the `tester` is restructured:
* the logic for running a single test is turned into a procedure
  (`runTest`)
* the tester no longer invokes itself -- `runTest` is called directly
  instead
* both tester-built and external runners use the `RunnerDesc` type
  for describing how they're invoked

In order to be used instead of the previous `expr` test runner, `phy`
now ignore VM modules without procedure when in `--runner` mode.

Finally, the old custom runners are removed and replaced with usage of
`phy`. The specs of some tests are adjusted to the slightly different
behaviour of `phy` (i.e., output formatting and arguments).

---

## Notes For Reviewers
* the follow-up to #61
* CI got quite a bit faster. The more passes there are, the larger the gains will be (over using a custom runner for each directory)


tools/tester.nim
```diff
@@ -265,39 +328,59 @@ if file.len == 0:
...
-        p.close()
 
   if success == total:
     echo "Success! Ran ", success, " tests"
   else:
     echo "Failure"
     quit(1)
 else:
+  # legacy mode:
+  var desc = RunnerDesc(exe: runner)
```
saem, line 371: IIUC a `var` is not required.

saem (approved):
A possible minor removal of a `var` local, otherwise LGTM.

tools/tester.nim
zerbina: Yep, it's not, and the `RunnerDesc` is also incomplete (it misses the argument specification).

PR #66, opened 2024-11-03
lang: add the logical `And` and `Or` operators

zerbina:
## Summary

Add the logical `And` and `Or` operators to the source language.

## Details

Both operators short-circuit evaluation of the second operand. For this
reason, dedicated expression kinds are used instead of built-in calls.

Since the second expression of both `And` and `Or` is evaluated
conditionally, it is analyzed within its own scope in order to prevent
declarations therein from being visible to the outside -- such locals
would be in an undefined state when the expression is not evaluated.

`And` is implemented as the statement version of `(If a b False)`; `Or`
is implemented as the statement version of `(If a True b)`.

### DIFF
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -422,6 +422,18 @@ proc fitExpr(c; e: sink Expr, target: SemType): Expr =
     result = Expr(stmts: e.stmts, expr: @[Node(kind: IntVal)],
                   typ: errorType())
 
+proc fitExprStrict(c; e: sink Expr, typ: SemType): Expr =
+  ## Makes sure expression `e` fits `typ` exactly, reporting an error and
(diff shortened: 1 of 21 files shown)


passes/source2il.nim
```diff
@@ -422,6 +422,18 @@ proc fitExpr(c; e: sink Expr, target: SemType): Expr =
     result = Expr(stmts: e.stmts, expr: @[Node(kind: IntVal)],
                   typ: errorType())
 
+proc fitExprStrict(c; e: sink Expr, typ: SemType): Expr =
```
saem, line 425: I guess the other option is we cast if it's a subtype, but this is probably for the best.

saem (approved):

PR #67, opened 2024-11-04
lang: make procedures first-class

zerbina:
## Summary

* add procedural types and procedural type constructors
* allow using procedures as values

## Details

.. to be written ..

---

## To-Do
- [ ] write the specification for procedural types and values
- [ ] implement procedural type constructors
- [ ] implement procedural values
- [ ] add tests
- [ ] write a proper commit message

### DIFF
--- a/common/vmexec.nim
+++ b/common/vmexec.nim
@@ -12,7 +12,7 @@ import
     utils
   ]
 
-import vm/vmtypes except tkVoid, tkInt, tkFloat
+import vm/vmtypes except tkVoid, tkInt, tkFloat, tkProc
 
 proc readInt(p: HostPointer, size: range[1..8]): int64 =
   copyMem(addr result, p, size)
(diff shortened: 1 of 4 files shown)


zerbina:
There's no dedicated `nil`-like value for `proc` types at the moment, but I think it'll be needed in the future. Maybe we've come up with a better solution by then, so I'm not in a rush to add a `nil`-like value for `proc` types right now.

An anonymous union such as `(UnionTy (UnitTy) (ProcTy ...))` (effectively an `Option` type) would work but has the problem of being larger than just a `ProcTy` (unless certain type combinations in unions are guaranteed to be packed, but this has its own set of issues).

Another thing to consider is whether a distinction should be made between a *procedure value* and procedure *pointer value* (similar to C, where a distinction is made between function types (e.g., `void (int)`) and pointer-to-function types (e.g., `void(*)(int)`). At the moment, all procedural values can only be (and are) pointers, and thus the distinction wouldn't be very useful, but this might change once metaprogramming enters the picture and one wants to pass around concrete procedures (and operate directly on them).

zerbina:
The test failures are caused by the bug(s) fixed by #68.

PR #70, opened 2024-11-16
add the `skully` program

zerbina:
## Summary

Skully (a portmanteau of "Skull" and "Phy") is a backend for NimSkull
that targets the `L25` language. It's combined with the NimSkull
frontend into a standalone program that takes a project module as input
and writes the generated `L25` code to a specified file.

Most NimSkull core language features are supported. The idea with
skully is to get access to large amounts of `L25` code resulting from
real-world code, for the purpose of testing and measuring the lowering
passes and VM.

## Details

The implementation is split into two parts:
* the actual code generator (`backend`) implementing the MIR to `L25`
  translation
* the program (`skully`) combining the NimSkull front- and mid-end with
  the Skully code generator

The `L25` IL is chosen as the target language because it's the language
most similar to the MIR.

Except for C interop, the `L25` is able to support the same language
features as NimSkull's C backend, and therefore `sem` is configured as
if it were run prior to the C backend. A non-trivial configuration is
used to make compilation possible:
* the target platform is hardcoded to be `any`, disabling most
  platform-specific code
* various internal conditional symbols are enabled to disable as much
  C-reliant code as possible
* a hook mechanism is used to replace importc'ed procedures with
  pure-NimSkull implementations or redirect them to dedicated VM host
  procedures
* platform-specific modules are replaced with skully-provided modules
  (only `os` at the moment, but the facility is generic and can
  redirect any module)

The VM host procedure specific to Skully are implemented into `phy`, so
that the `phy` program can run `skully`-produced code.

### Code Generator

The code generator performs a translation from MIR code to `L25` code,
attempting to replicate all lowerings NimSkull's C code generator
performs, meaning that the behaviour of the generated code *should* be
identical.

It's hooked into the compilation process like the native backends are:
via the `backends.process` iterator.

RTTIv1 generation is complicated and only needed for `deepCopy` and
the `typeinfo`, therefore it's not implemented by Skully; only the
RTTIv2 is, as it's necessary for ORC and `of` support.

### Platform Support

Only the OS procedures necessary for compiling the `phy` program are
implemented at the moment.

### Tests

For making sure Skully works at least somewhat, a CI job is added that
compiles `phy` to `L25` code and then executes said `L25` code to build
and run an more complex test from the `expr` tests.

---

## To-Do
- [x] implement set constructors
- [x] implement the various set magics
- [x] implement the remaining non-set magics
- [x] implement RTTI generation
- [x] implement the main procedure
- [x] implement proper overrides for the I/O procedures
- [x] merge the main branch once the pass rework is done and merged
- [x] finish the incremental procedure code-gen
- [x] implement `getExecArgs`
- [x] use `--gc:orc`
- [x] register `skully` with `koch`
- [x] add a test for making sure that `skully` can compile a working `phy` program
- [x] cleanup the code


zerbina:
Here are some pass timings gathered by using a `-d:release` `skully` to compile `tools/repl.nim` (~4 MiB of `L25` code).

| Pass | Time Taken (in ms) |
| -----|----------- |
| `pass25` | 4,579 |
| `pass4`  | 121,116 |
| `pass3`  | n/a |
| `pass1`  | n/a |
| `pass0`  | n/a |

Note that the current skully is not able to run all passes, as the code passed to `pass25` is - strictly speaking - not valid (because it uses aggregate parameters and return values).

In the current pass pipeline, `pass25` and `pass4` are the most complex and time intensive ones. Nonetheless, 2 minutes (!) to run the first two passes is completely unacceptable, of course.

These long run times are mostly due to the missing moves I've already mentioned [here](https://github.com/nim-works/phy/pull/52#issuecomment-2395629664). Fortunately, the fix is very simple, and applying it, the timings I get are:

| Pass | Time Taken (in ms) |
| -----|----------- |
| `pass25` | 388 |
| `pass4`  | 5,444 |
| `pass3`  | n/a |
| `pass1`  | n/a |
| `pass0`  | n/a |

which is a lot more reasonable, but still way too slow.

For comparison, here are the timings for the new passes (after making the necessary, small adjustments to `skully` in order for it to work with the new passes):

| Pass | Time Taken (in ms) |
| -----|----------- |
| `pass25` | 13 |
| flatten paths | 30 |
| aggregate params | 26 |
| aggregates to blob | 39 |
| locals to blob | 13 |
| legalize blob ops | 16 |
| stack alloc | 22 |
| inline types | 10 |
| `pass0` | 12 |

The commit used is https://github.com/nim-works/phy/pull/85/commits/c9c771c9b8bc9e38563d575cf42cb3acbfa598f7. Beyond being *significantly* faster, the intermediate pass results are also *much* smaller.

zerbina:
`skully` is almost able to compile `phy`, only an implementation of `getExecArgs` is still missing. There are also a few NimSkull bugs that need to be fixed in the upstream.

zerbina:
Okay, `skully` is now able to compile the `phy` program into `L25` code, which can then be run via the native `phy` executable (refer to the test I added to `build_and_test.yml`).

This wraps up the main part of the work. Implementing some small leftovers and refactoring the code a bit should be all that there's left to do.

zerbina:
I've cleaned up the code a little, shuffling some code around and rewording some comments. There's probably a lot more that could be done, but given its standalone nature (i.e., `skully` not being part of the core), I personally think the code in its current shape is fine.

`skully` is - more or less - a fully-fledged compiler backend on the level of the NimSkull-native C code generator, with all the complexity that currently entails: the MIR code is not adequately lowered, thus leaving a lot of work to the downstream processing (i.e., the code generator).

zerbina:
Hm, it looks like some upstream changes uncovered some code generator bug. I suspect that the fix itself is going to be relatively simple and low in impact, so I'll mark the PR as ready-for-review already.

@saem: For the review, I think it'll be easiest to review the final diff rather than going commit by commit, as there's was a lot of code churn. I'd recommend the following order of looking over the changes:
1. `skully.nim`
2. `backend.nim`
3. `host_impl.nim` changes
4. `patch` directory and the CI changes  

There are most likely some latent bugs in there, and if you come across something that looks off to you, please bring it to my attention. Nonetheless, I consider `skully` to be stable enough for its intended purpose.

zerbina:
As expected, the fix was very simple. The moving-around of some code had messed up indentation, resulting in trees not adhering to the `L25` grammar.

saem (approved):

PR #72, opened 2024-11-22
lang: add `While` loops

zerbina:
## Summary

Add the `While` loop construct to the source language. A `While` loop
repeatedly evaluates a given unit expression as long as the given
condition evaluates to true.

## Details

A `While` loop is implemented as a `Loop` + conditional `Break` in the
target IL.

At the moment, a `(While (Ident "true") ...)` is always typed as
`void`, as - without a `Break`-like statement - control-flow can never
reach the code immediately following the loop. Due to the current lack
of early constant folding, detection of such loops needs to be syntax-
based, and thus more complex cases (such as `(Exprs (Ident "true"))`)
are *not* typed as `void`.

### DIFF
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -227,6 +227,10 @@ The type of a `While` expression depends on the `cond` expression. If `cond`
 is the built-in `true` literal `(Ident "true")`, then the expression is of
 type `void`, otherwise it is of type `unit`.
 
+> TODO: once constant expression evaluation is specified, consider changing the
+>       rules such that a `While` is of type `void` when the is a constant
+>       expression that evaluates to true
+
 **Expression kind**: r-value
 **Uses**: `cond` and `body`
 
--- a/spec/spec.md
+++ b/spec/spec.md
@@ -209,6 +209,27 @@ The type of the `If` expression is the common type between `A` and `B`.
 **Expression kind**: r-value
 **Uses**: `cond`, `body`, and - if present - `else`
 
+#### `While`
+
+```grammar
+expr += (While cond:<expr> body:<expr>)
+```
+
+Repeatedly evaluates `body`, as long as `cond` evaluates to `true`. Both
+`body` and `cond` are part of a new scope.
+
+Let `C` be the type of `cond` and let `T` be the type of `body`. An error is
+reported if:
+* `C` is not `bool`
+* `T` is not `unit`
+
+The type of a `While` expression depends on the `cond` expression. If `cond`
+is the built-in `true` literal `(Ident "true")`, then the expression is of
+type `void`, otherwise it is of type `unit`.
+
+**Expression kind**: r-value
+**Uses**: `cond` and `body`
+
 #### `Return`
 
 ```grammar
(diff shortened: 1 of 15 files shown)


zerbina:
`While` loops themself don't need `Continue` and `Break` source language statements to be usable, so I think it's better to implement them in a separate PR.

tests/expr/t15_while_true.test
```diff
@@ -0,0 +1,10 @@
+discard """
+  description: "
+    A `(While (Ident true))` loop never exits normally, and is thus of type
+    void
+  "
+  arguments: "c"
```
saem, line 6: am I missing something, why is the `arguments` necessary?

spec/spec.md
```diff
@@ -209,6 +209,27 @@ The type of the `If` expression is the common type between `A` and `B`.
...
+expr += (While cond:<expr> body:<expr>)
+```
+
+Repeatedly evaluates `body`, as long as `cond` evaluates to `true`. Both
+`body` and `cond` are part of a new scope.
+
+Let `C` be the type of `cond` and let `T` be the type of `body`. An error is
+reported if:
+* `C` is not `bool`
+* `T` is not `unit`
```
saem, line 224: presuming that a hypothetical `quit` procedure would be `() -> void`, then wouldn't this definition be problematic, as one couldn't put `quit` as the `body`?

N.B.: I'm ok punting on this and figuring it out later

spec/spec.md
```diff
@@ -209,6 +209,27 @@ The type of the `If` expression is the common type between `A` and `B`.
...
+
+The type of a `While` expression depends on the `cond` expression. If `cond`
+is the built-in `true` literal `(Ident "true")`, then the expression is of
+type `void`, otherwise it is of type `unit`.
```
saem, line 228: should this document the expected behaviour, that if the condition evaluates to true after constant expression evaluation then the loop is `void`, with a note saying the implementation doesn't yet do that?

saem (approved):
Looks good!

A few minor questions, but they're all soft wrt requiring action.

tests/expr/t15_while_true.test
zerbina: Ah, I'll add a comment explaining why. The while loop repeats forever, which requires not running the test, otherwise the runner would never return.

spec/spec.md
zerbina: Hm, you're right. I'm not sure what I was thinking when disallowing `void` here. Not only a hypothetical `quit` cannot be put as the `body`, a `Return` or `Unreachable` cannot be put there too.

I'll adjust the rules.

spec/spec.md
zerbina: A note mentioning the direction is a good idea, though I would leave the mentioning implementation out of it, so that the specification stays independent from the implementation.

Unrelated to the note, but macros and constant expressions are going to cause us some trouble, semantics wise. If compile-time evaluation is only partially defined (like it is in NimSkull), then the static semantics of the language become implementation defined, which I consider quite problematic. Having *all* dynamic behaviour be well-defined is going to be a challenge, I believe.

PR #73, opened 2024-11-23
support standalone numbers and symbols in source code

zerbina:
## Summary

In S-expressions representing source language code, symbols and numbers
can now appear anywhere. If not appearing in a special position,
they're parsed as `(Ident x)` and `(IntVal x)`/`(FloatVal x)`,
respectively.

For example, `(Asgn (Ident "x") (IntVal 1))` can now be written as
`(Asgn x 1)`.

## Details

Standalone symbols and numbers can be unambiguously parsed as `Ident`
and `IntVal`/`FloatVal` nodes. The aim is to make writing source
language code easier (until a non-S-expression syntax is developed).

In terms of implementation, `tree.fromSexp` now calls the generic
`fromSexp` and `fromSexpSym` procedures provided for the node kind
enum.

For `spec_source` (the source language tree representation), standalone
symbols parse as `Ident`, floating-point numbers as `FloatVal`, and
integer numbers as `IntVal` (instead of `Immediate`). The `Immediate`
node kind is not used anywhere else, so it's removed.

The enum for the IL node kinds does not support standalone S-expression
symbols (because there's no ident-like node kind), but it does support
standalone floating-pointer numbers now (which are parsed as `FloatVal`
nodes).

---

## Notes For Reviewers
* not having to write out `Ident`/`IntVal`/`FloatVal` makes the S-expression representation feel much more like a "real" language
* the plan is still to eventually introduce a non-S-expression syntax

### DIFF
--- a/passes/spec.nim
+++ b/passes/spec.nim
@@ -69,5 +69,14 @@ proc toSexp*(tree: PackedTree[NodeKind], idx: NodeIndex,
   of Global:    sexp([newSSymbol("Global"), sexp n.val.int])
   else:         unreachable()
 
-proc fromSexp*(i: BiggestInt, _: typedesc[NodeKind]): TreeNode[NodeKind] =
+proc fromSexp*(tree: var PackedTree[NodeKind], i: BiggestInt
+              ): TreeNode[NodeKind] =
   TreeNode[NodeKind](kind: Immediate, val: i.uint32)
+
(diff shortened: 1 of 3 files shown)


saem (approved):
this is much easier to type and we can still use `(ident "escape me")` for escaping/stropping.

PR #75, opened 2024-11-27
vm: use a simpler type representation

zerbina:
## Summary

Simplify the type representation used by the VM in order to reduce the
size of types (in memory) and cost of validation.

## Details

Instead of having a list of types and a list with parameters (for
signature types), theres now a list of signatures and a list of
parameter types. `TypeId`s can only refer to signature types now.

This has a multiple benefits:
* there's no more requirement that the first element in the type list
  is a `tkVoid` type
* the type environment takes up less space in memory (and on disk)
* looking up a parameter type requires one less indirection

The VM validation layer is updated accordingly. For the assembler, the
type directive is changed such that only signature types are allowed,
whose syntax is now `(T*) -> T`, which should make it clearer what the
syntax represents (compared to the previous `(Proc ...)`.

For the tests, the expected output of `pass0` tests as well as the `vm`
tests are updated to the changed assembler syntax.

There are some changes relevant to embedders of the VM:
* the `initTypeEnv` procedure is removed, as an empty `TypeEnv` is now
  valid
* `YieldReason.typ` is a `TypeKind` instead of `TypeId` now 

---

## Notes For Reviewers
* a preparation for adding VM modules

### DIFF
--- a/common/vmexec.nim
+++ b/common/vmexec.nim
@@ -166,8 +166,8 @@ proc run*(env: var VmEnv, prc: ProcIndex): string =
   result = "(" & substr($res.kind, 3)
   case res.kind
   of yrkDone:
-    case env.types[res.typ].kind
-    of vmtypes.TypeKind.tkVoid, vmtypes.TypeKind.tkProc, tkForeign:
+    case res.typ
+    of vmtypes.TypeKind.tkVoid, tkForeign:
       discard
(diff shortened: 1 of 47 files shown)


saem (approved):
that encoding is really nice, even the little things like being able to use the init/zero value.

PR #76, opened 2024-11-27
lang0: rename blit operation to `Blit`

zerbina:
## Summary

Instead of `Copy`, the memcopy/blit operation is now named `Blit`,
making it clearer what the operation does.

## Details

Both the copy and memcopy operation used the name `Copy`, which -
besides being confusing - made searching for either of them using just
the node kind impossible (the shape had to be considered for
discriminating between them).

---

## Notes For Reviewers
* this is a split-out from the pass rework


saem (approved):
a blit is a move or copy of a data block, but people typically think move, not entirely convinced it's clearer, but it is _far_ more searchable, so I get the motivation.

PR #79, opened 2024-12-01
lang: add built-in dynamic arrays

zerbina:
## Summary

Add run-time sized arrays with value semantics to the source language.
They're of type `SeqTy`, are constructed with `Seq` expressions, are
concatenated with `concat`, and have their dynamic length queried with
`len`.

## Details

Dynamic arrays are implemented as a length + pointer tuple, with the
pointer pointing to a dynamically allocated payload storing a capacity
(an integer) and an array of elements. Each use of a dynamic array
creates a full copy, which means creating a duplicate of the payload
and all elements within.

Growing a dynamic array and appending to it are implemented via
always-present type-erased L30 procedures. The copy procedure for types
needing one (queried via the new `isTriviallyCopyable` procedure) are
synthesized and cached when first needed. Calls to the copy procedure
are injected by the `use` procedure, which replaces all `inline`
usage in contexts where the specification describes a `copy` taking
place.

To support multiple procedures being constructed concurrently (needed
for ad-hoc type-attached operator synthesis), IL procedure definitions
are now first added to an intermediate seq before being added to the IL
module upon closing the analysis context.

Dynamic array values could and should be destructed when no longer
needed, but since destruction cannot be observed via in-language means
at the moment, it cannot be tested properly and is thus not
implemented.

Finally, `source2il` now keeps track of the most recently added user-
declared procedure's index in the `entry` field. With the synthesized
operators, the last procedure in the module's list is no longer
guaranteed to be a user-declared one, so `phy` and `repl` module use
the `entry` field to know which procedure to use as the entry
procedure.

---

## To-Do
* [x] implement the internal allocator (`alloc`, `realloc`, and `dealloc`)
* [x] implement the internal `prepareAdd` routine
* [x] implement destructor generation and destructor call injection
* [x] implement copy/clone generation and copy call injection
* [x] add tests for `Seq` and `SeqTy`
* [x] add tests for `add` and `concat`


zerbina:
To keep the implementation simple(r) for now, and in order to not further complicate the still-in-progress pass rework, there's going to be no separate copy-to-move optimization pass or destructor injection pass (like there is in the NimSkull compiler).

Instead, all destructors and copy calls will be injected right away, as part of `source2il`. As a consequence, all usages of `seq` l-values outside of `len`, `clear`, and `add` calls require a full copy to be made. Once the pass rework is done, a copy-to-move optimization pass and the associated intermediate languages can be added, which will address this inefficiency.

zerbina:
I'm going to split the parts related to the allocator into their own PR, as they're quite involved and not dependent on the `seq` implementation.

zerbina:
I've changed plans a bit. `add` and `clear` would both require a custom evaluation order, making them have - in my opinion - too much overhead in terms of definition effort. I've therefore removed both and added `concat` in `add`'s place (`concat` returns a new sequence instead of appending to an existing one).

Cleanup of locals is also kept simple, ignoring any temporaries spawned. Given that the allocator doesn't reuse freed memory and that there are no copy/destroy hooks, this cannot be observed and is therefore not a problem needing to be solved at this time.

saem (approved):

PR #82, opened 2024-12-10
vm: fix incorrect stack state after `IndCall`

zerbina:
## Summary

* fix the operand stack being in an incorrect state after `IndCall`
* fix the assembler interpreting all numbers as hexadecimal numbers
* change the assembler to require a type name for `IndCall`

## Details

For convenience of the VM, the callee operand for `IndCall` must be on
top of the stack. It has to be popped before entering the procedure, so
that the arguments are on top of the stack, but it wasn't.

There were also two issues/bugs with the assembler that are now fixed:

* number parsing tries interpreting numbers as decimal first, so that
  numbers aren't always interpreted as hexadecimal (since `parseHex`
  also supports numbers without a prefix)

* `IndCall` parsing required a type ID rather than a type name, which
  is inconsistent with all other places expecting a type. A type name
  is now expected

Finally, tests for the `IndCall` operation are added to ensure that it
works correctly.

---

## Notes For Reviewers
* this is a bug that was resurfaced by #70


zerbina:
It'd be better if the callee operand comes after the procedure arguments, so that the latter are on top of the stack, but that would require shifting the stack operands to the left by one element, and I want to first do a performance test with [skully](https://github.com/nim-works/phy/pull/70) (which is blocked by the bug this PR fixes) in order to make sure that the impact is acceptable.

saem (commented):
LGTM

PR #85, opened 2024-12-13
rework the passes and intermediate languages

zerbina:
Redesign large parts of the intermediate languages, fixing some mistakes made with their first iteration.

Some terminology is changed (e.g., continuation -> basic block), transformations are redistributed between passes, new passes are introduced, and the languages are simplified. The most significant simplification is the temporary removal of the SSA form and associated transformation. It results in simpler ILs, much smaller IL code sizes, and much faster passes, at the cost of a major regression in quality of the generated VM bytecode.

In addition, the following important features are now supported by the `L30` IL:
* alignment requirements for aggregate types
* locals of scalar type can have their address taken
* procedure parameter and return values can be of aggregate type

---

## To-Do
* [ ] implement the new passes
* [ ] adjust the existing passes where necessary (`pass0` and `pass25`)
* [ ] change `source2il` to produce valid `L30` code (according to the new specification)
* [ ] write tests for the new passes
* [ ] remove the obsolete tests and passes
* [ ] write a proper commit message

## Notes For Reviewers
* this PR is the successor to #64

### DIFF
--- a/passes/lang0.md
+++ b/passes/lang0.md
@@ -5,20 +5,14 @@ abstractions over control-flow, stack allocation, reinterpretation and
 conversions, and arithmetic operations, but generally stays close to VM
 bytecode.
 
-Procedures are split into continuations. They need to be ordered such
-that all continuation exits (except for `Loop`) describe forward
-jumps. One (and only one) continuation must be the "return" continuation.
-
 ```grammar
 numtype  ::= (Int size:<int>)
           |  (UInt size:<int>)
           |  (Float size:<int>)
 typedesc ::= (ProcTy (Void) <type>*)
           |  (ProcTy <type>+)
-          |  <numtype>
 type_id  ::= (Type <int>)
-type     ::= <type_id>
-          |  <numtype>
+type     ::= <numtype>
 ```
 
 A type can be either *inlined* (anonymous) or *identified*. An *identified*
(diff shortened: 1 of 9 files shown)


zerbina:
The reworked passes are not perfect and by no means final. With the second attempt, a major goal of mine was to speed up future development of the passes, hence me making simplicity a priority and removing any form of SSA transformation for now (it will most likely be reintroduced at some point in the future).

I've tested the new ILs extensively with [skully](https://github.com/nim-works/phy/pull/70) (using a proof-of-concept implementation) in order to make sure that they're powerful enough to work as the target language for NimSkull's MIR and also that it's actually possible to implement the necessary passes in an efficient and reasonably simple manner (something where the previous iteration fell short) -- which they showed to be.

zerbina:
To facilitate the new passes - or at least make their implementation cleaner -, I've extended the `ChangeSet` API with "virtual trees", which finally expose the ability to create stacked changes in a reasonably ergonomic manner.

The `Cursor` API from the predecessor PR was an earlier attempt at this, which - while quite efficient - was made infeasible to use for some transformations (like the aggregate parameter lowering) by its requirement to process the input top-down outside-to-inside.

Another major improvement is the new `buildTree` macro. It took me some time to convince myself that a macro is warranted here, but in the end, I think using one is/was the right choice. The macro is used like this:
```nim
bu.buildTree:
  tree(Asgn, node(Local, ...), tree(Copy, node(Local, ...)))
```
allowing one to express trees using an almost S-expression-like syntax. Beyond being easier to read/write than `subTree`/`add`, it's also more efficient, as the memory for all nodes in the constructed tree can be allocated upfront and in one go.

The macro itself is quite simple, only using `quote` and the low-level `NimNode` API and leaving almost all of the input validation to the compiler, so it shouldn't cause too much trouble in the future.

zerbina:
The aggregate parameter lowering pass is the most complex one out of the new passes. Most of it is boilerplate however, and a lot of shortcuts are taken, resulting in rather low-quality output. Reintroduction of an SSA form plus the introduction is going to help here, and the pass could also limit hoisting to impure expression (and then only when the materialized r-value expression is impure itself).

phy/phy.nim
```diff
@@ -198,22 +208,34 @@ proc compile(tree: var PackedTree[spec.NodeKind], source, target: Language) =
       assert false, "cannot be handled here: " & $current
     of lang1:
       syntaxCheck(tree, lang1_checks, module)
-      # TODO: don't hardcode the pointer size
-      tree = tree.apply(pass1.lower(tree, 8))
+      tree = tree.apply(pass_inlineTypes.lower(tree))
       current = lang0
+    of lang2:
+      syntaxCheck(tree, lang2_checks, module)
+      # TODO: don't hardcode the pointer size
```
saem, line 215: does it make sense to lift this out as a const for now, so it's easier to change to a variable that's a bit more dynamic taking configuration into account?

saem (approved):
one minor item, but it looks good when I ran through it

phy/phy.nim
zerbina: Yep, good call.

PR #86, opened 2024-12-16
source2il: use a tree-based IR

zerbina:
## Summary

Construct IL expressions and statements using a nested tree-based IR,
replacing the usage of `Builder` and making tree construction easier to
read and write.

## Details

`source2il` has to perform various lowering intermixed with analysis at
the moment, which makes the `Builder` API cumbersome and noisy to use
for constructing the trees.

Using a tree-based IR - where each node stores all its transitive child
nodes - makes construction much easier, at the cost of more allocator
activity (since each node with subnodes allocates its own `seq`). The
IR also handles some additional details, like packing numeric values
and wrapping lvalue expression, further simplifying construction.

The tree-based IR only covers the statements and expressions, not
things like procedures and types. Various procedures are provided for
constructing common tree shapes (assignments, if statements, etc.).

In order to prevent bugs (due to shared ownership) and improve memory
locality, no `ref`s are used. Copying a node therefore means creating a
full copy of the subtree, which makes it important for performance that
`sink` is used when passing nodes to construction procedures.

---

## Notes For Reviewers
* meant to assist with #79
* once lowering has largely moved out of `source2il`, the IR should be removed again

### DIFF
--- a/passes/ir.nim
+++ b/passes/ir.nim
@@ -1,0 +1,197 @@
+## Implements an IR for the highest-level IL that's not flat. It's meant as a
+## helper for ``source2il``, to make tree construction easier and less prone
+## to mistakes. The IR is meant to be a temporary measure until more lowering
+## is delegated to separate passes.
+##
+## Some common things, like wrapping lvalue expression in a `Copy` operation
+## when appearing in a usage context, are handled during to-packed-tree
+## translation.
+
+import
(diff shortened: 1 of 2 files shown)


saem (approved):
the scratch pad style and then compact approach looks pretty good, at least from a compactness point of view.

PR #88, opened 2024-12-17
source2il: show proper types in error messages

zerbina:
## Summary

Use properly rendered types in source language diagnostic messages
instead of showing just the raw name of the type kind enum value.

## Details

* move the type rendering from `vmexec` into its own module
* render the full types when formatting error messages in `source2il`
* document why rendering for `tkError` is not implemented

---

## Notes For Reviewers
* a small QoL improvement
* should be merged prior to #79 in order to prevent merge conflicts


saem (approved):
Sorry for the delay, but I see how this was a detour from the main thrust of work

PR #89, opened 2024-12-18
lang: add array indexing

zerbina:
## Summary

Add the `At` operator to the source language, for accessing elements of
arrays. All array access is checked, with out of bounds access
resulting in a panic.

## Details

* add the static and dynamic semantics for `At` to the specification
  * the interal-only `With` abstract syntax is introduced, to make
    defining the reduction of field/element assignments easier 
  * some small problems (incorrect notation, unintended meaning) with
    unrelated rules are fixed
  * the field access semantics is modified such that it matches the
    old one (in theory, no specification tests exist yet)

* implement the `At` syntax in `source2il`. Computing the element's
  address and checking the index needs to be an atomic expression, and
  thus the combined access/check is done via a call to procedure, which
  is synthesized for a `Seq` type upon first `At` use

* add tests for `At` and improve the existing `t17_seq_copy.test`

### Evaluation Order

The index operand may be an arbitrary effectful expressions, and
evaluating it can thus replace the dynamic array value in the location
that appears in the array operand position.

To prevent use-after-free errors without copying, the new evaluation
order for lvalue expressions is not strictly left to right. Instead,
all non-lvalue operands in the lvalue expression are evaluated first,
and only afterwards is the lvalue expression evaluated.

---

## To-Do
* [x] merge the main branch once the `seq` type is added to the language
* [x] implement `At`
* [x] write tests for `At`


zerbina:
The evaluation order may seem a bit odd at first, but it being the way it is is deliberate. In order to understand why, consider the following NimSkull code:

```nim
var x = @[1]
discard x[(x = @[]; 0)]
```

With a strict left-to-right evaluation order, the `x` would be evaluated first. Since the `[]` operator in NimSkull itself never copies the array operand, the value `x` evaluates to would be *non-owning*. Evaluating the index expression would then destroy the value currently stored in `x`. The value `x` evaluated to earlier would still refer to the previous payload and length, however, and thus the access would result in a use-after-free.

In practice, in this very case, evaluation order is *not* left-to-right in NimSkull. To my knowledge, this is not documented anywhere, but it's visible in the MIR code:
```
scope:
  def x: seq[int]
  def _2: seq[int] = @[]
  x = move _2
  chckIndex(arg x, arg 0) -> [L0]
  discard x[0]
  destroy x
  goto [L1]
  finally (L0):
    destroy x
    continue [Resume]
  L1:
```
When nesting the seq access, evaluation falls back to being left to right, but only when index checks are enabled (!).

This same problem also affects the source language, as said language too allows for index operands with side-effects. There are four possible solutions:

1. disallow index operands with side-effects
2. copy the array operand (only needed for dynamic arrays)
3. disallow such problematic mutations via a borrow checker. `(At x y)` would borrow `x` until the full enclosing lvalue expression is evaluated
4. don't use a left-to-right evaluation order

In my opinion, option 1 is too restrictive and option 2 would introduce a performance trap, where seemingly innocuous code performs a (potentially very costly) copy. Option 3 could be a candidate to consider in the future, but it requires a borrow checker, which is something not in reach at the moment.

Therefore, only option 4 remains. I've chosen an evaluation order that I think is also somewhat intuitive: all non-lvalue operands in an lvalue expression are evaluated with the usual left to right, inside to outside order first, and then lvalue operands are evaluated.

In practice, this works as follows (using NimSkull syntax for the sake of readability):
```nim
var s: seq[seq[int]]
echo s[0][(s = @[@[1]]; 0)]
# echoes 1

s = @[@[0]]
s[s[0][0]][3] = (s = @[@[1, 2, 3, 0]]; 4)
# step by step evaluation (excluding uninteresting ones):
s[0][3] = (s = @[@[1, 2, 3, 0]]; 4) # 1; s == @[@[0]]
s[0][3] = 4                         # 2; s == @[@[1, 2, 3, 0]]
-                                   # 3; s == @[@[1, 2, 3, 4]]
```

saem (approved):

PR #91, opened 2024-12-20
assembler: add the `.import` directive

zerbina:
## Summary

Add the `.import` directive to the assembler, which makes it possible
to express imported/foreign procedures in assembler code.

## Details

The import name uses double-quotes, even though the currently
supported character set doesn't require quotation. This is only
for reasons of style and readability, to highlight that the name
isn't a normal identifier.

---

## Notes For Reviewers

A blocker for #70. The upcoming test for skully needs to disassemble and reassemble a VM module, and imported/foreign procedures don't roundtrip at the moment.

### DIFF
--- a/tests/vm/runner.nim
+++ b/tests/vm/runner.nim
@@ -17,7 +17,7 @@ import
   ]
 
 proc test(env: var VmEnv, args: openArray[Value], _: RootRef): CallbackResult =
-  ## A VM callback.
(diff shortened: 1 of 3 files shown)


saem (approved):
Minor suggestion for clarifying the test vm callback comment.

PS. I updated the PR description to call it `double-quote` instead of `double-tick quote`, as `tick` is evocative of '`'

PR #92, opened 2024-12-20
source2il: implement a simple allocator

zerbina:
## Summary

Implement an allocator for internal use by the `source2il`, for
implementing language features that require heap locations (e.g.,
dynamic arrays).

## Details

The allocator is a simple bump-pointer allocator backed by a static
memory region. It consists of three procedures: `alloc`, `dealloc`, and
`realloc`. Since the source language doesn't support the features
necessary for implementing the allocator, the aforementioned procedures
are implemented in `L30` code provided via inline strings.

`dealloc` is currently a no-op, meaning that memory is never freed.
Upon exhausting the available heap memory, an exception is raised,
which currently results in undefined behaviour (as `CheckedCall` isn't
used for call expressions).

To reduce implementation complexity, the allocator procedures are
always added to the IL module, even if not used. Since this interferes
with the previous entry procedure selection (the last procedure in the
IL module was picked), all user-defined procedures are now marked as
exported, and for programs originating from source-language code, the
last *exported* procedure is now picked to be the entry procedure.

Finally, to be able to configure the guest memory from the outside via
modules, instead of it being hardcoded, the `phy` program now retrieves
the stack size, stack start, and total amount of memory from globals
with the interface names `stack_size`, `stack_start`, and
`total_memory`, respectively. This also means that the VM module is no
longer linked when not evaluating the program.

---

## To-Do
* [x] implement exports; needed for designating a main procedure (separate PR)
* [x] update the stack setup to use the size and position provided by the globals

## Notes For Reviewers
* split out from and a prerequisite of #79
* the memory configuration through exported globals is also needed for #70
* the allocator is not exposed to the source language, so it cannot be tested as part of this PR

### DIFF
--- a/common/vmexec.nim
+++ b/common/vmexec.nim
@@ -2,6 +2,9 @@
 ## runners.
 
 import
+  std/[
+    options
+  ],
   phy/[
     type_rendering,
     types
@@ -9,11 +12,19 @@ import
   vm/[
     vmalloc,
     vmenv,
+    vmmodules,
     vm,
     utils
   ]
 
 import vm/vmtypes except tkVoid, tkInt, tkFloat, tkProc
+
+type
+  MemoryConfig* = object
+    ## The guest memory configuration.
+    total*: uint
+    stackStart*: uint
+    stackSize*: uint
 
 proc readInt(p: HostPointer, size: range[1..8]): int64 =
   copyMem(addr result, p, size)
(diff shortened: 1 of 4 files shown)


zerbina:
Further down the line, I think that the module(s) should provide the guest memory configuration (maximum and initial size) through some dedicated facility, similar to how WASM does it. For now, the (not yet implemented) approach of providing the stack and memory configuration via exported globals works well enough.

common/vmexec.nim
```diff
@@ -83,17 +94,47 @@ proc valueToString(env: var VmEnv, a: VirtualAddr, typ: SemType): string =
   of tkVoid, tkError:
     unreachable()
 
-proc run*(env: var VmEnv, prc: ProcIndex, typ: SemType): string =
+proc readMemConfig*(m: VmModule): Option[MemoryConfig] =
+  ## Reads the guest memory configuration from `m` by looking for the
+  ## `stack_start`, `stack_size`, and `total_memory` global variables.
```
saem, line 99: I don't know how long this is going to be the case, I suspect the interface will change sooner than later, otherwise I'd suggest putting these behind an aggregate if possible to reduce name collisions

tools/repl.nim
```diff
@@ -154,11 +156,19 @@ proc process(ctx: var ModuleCtx, reporter: Reporter,
       echo "validation failure"
       return
 
-    var env = initVm(1024, 1024 * 1024)
+    var mem: MemoryConfig
+    if (let v = readMemConfig(module); v.isSome):
+      mem = v.unsafeGet
+    else:
+      # if reached, there's a bug in either source2il or the lowering passes
+      unreachable()
```
saem, line 164:
```suggestion
      unreachable:
        "failed to read MemoryConfig: bug in either source2il or the lowering passes"
```

soft suggestion, just slightly nicer to debug

saem (approved):
a couple of very minor comments, feel free to ignore

common/vmexec.nim
zerbina: There cannot be any name collisions, as there is no way for the source language to create exported global variables with a user-defined name. Putting them in an aggregate is also not possible, because the `L30` language only supports global variables of numeric types.

I could use a prefix for the names (e.g., `internal.stack_start`), but I think it's more likely that the approach of providing the guest memory configuration via exported globals is gone by the time user-defined globals are a thing.

saem: I forgot that `L30` doesn't support aggregates. I don't think it's worth making any further changes, thanks for providing more context.

tools/repl.nim
zerbina: I've slightly shortened the string; a pass bug is a lot less likely (due to their separate tests) than a `source2il` one.

PR #94, opened 2024-12-21
vmmodules: support exported procs/globals

zerbina:
## Summary

* add support for exported procedures and globals to VM modules
* implement procedure import resolution for `vmmodules.link`

## Details

Procedures and global variables can now be marked as exported, by
adding them to the `export` table; the assembler provides this
functionality through the new `.export` directive. The same global
variable or procedure can be exported multiple times.

Exported procedures are used to resolve procedure imports, which
the `pkCallback` entries are repurposed to represent prior to
linking. The host procedure table is now only tried when a
procedure import doesn't resolve to an exported procedure; if
that fails too, the procedure is turned into a stub.

Global variables cannot be imported at the moment, but being able
to export them is still useful for looking them up from the
outside via an interface name.

Finally, since procedure stubs serve no purpose prior to linking, the
validator now reports an errors on  encountering them.

### Misc

The assembler was left in an invalid state when parsing the interface
name for the `.import` directive failed. This fixed by only modifying
the assembler after parsing the interface name.
 
---

## Note For Reviewers

This PR further develops the VM module support, which provides the foundation for the later IL / source-language module support. In the short-term, exports are required for making globals and procedures available to `phy`'s VM runner, which is needed by both #70 and #92.

### DIFF
--- a/passes/pass0.nim
+++ b/passes/pass0.nim
@@ -660,8 +660,8 @@ proc translate*(module: PackedTree[NodeKind]): VmModule =
       result.ehTable.add prc.ehTable
     else:
       # must be a host procedure
-      result.host.add module.getString(module.child(def, 1))
+      result.names.add module.getString(module.child(def, 1))
 
       result.procs.add ProcHeader(kind: pkCallback,
                                   typ: module[def, 0].typ)
-      result.procs[i].code.a = result.host.high.uint32
+      result.procs[i].code.a = result.names.high.uint32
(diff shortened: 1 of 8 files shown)


zerbina:
This PR is ready for review, but merging should wait until after #91, so that some tests for resolving the imports can be added.

zerbina:
There was a small issue with the assembler being left in an invalid state after a parsing error, affecting both the `.import` and `.export` directive. I've fixed it and updated the PR message.

saem (approved):
A soft suggestion, otherwise LGTM

PR #97, opened 2024-12-23
lang0: support exporting procedures/globals

zerbina:
## Summary

Add export lists - for exporting procedures and globals - to the `L0`
and all ILs extending it. The concept of *foreign procedures* is
changed into the more general *imported procedure* one.

## Details

A `Module` sub-tree may now have an optional trailing `List` sub-tree
containing a list of `Export` trees, which are translated directly into
VM module exports.

The `Foreign` node kind is renamed to `Import`. No further change is
required, as `L0` `Foreign` procedures are translated to `pkCallback`
procedure headers, which already represent procedure imports.

In addition, the disassembler output for global variables not being
valid assembler code is fixed.

---

## Notes For Reviewers
* further progress towards multi-module support
* support for exported globals in the ILs is a prerequisite for the allocator

### DIFF
--- a/tests/pass0/t05_export_global.test
+++ b/tests/pass0/t05_export_global.test
@@ -1,5 +1,5 @@
 discard """
-  output: "(Done 100)"
+  output: "(Done 200)"
 """
 (TypeDefs
   (ProcTy (Int 8)))
@@ -9,6 +9,6 @@ discard """
   (ProcDef (Type 0) 0 (Locals)
     (List
       (Block (Params)
-        (Return (IntVal 100))))))
+        (Return (IntVal 200))))))
 (List
   (Export "exported" (Global 0)))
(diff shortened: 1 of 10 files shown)


tests/pass0/t05_export_global.test
```diff
@@ -0,0 +1,14 @@
+discard """
+  output: "(Done 100)"
+"""
+(TypeDefs
+  (ProcTy (Int 8)))
+(GlobalDefs
+  (GlobalDef (Int 8) (IntVal 100)))
```
saem, line 7:
```suggestion
  (GlobalDef (Int 8) (IntVal 200)))
```

I know we're not outputting this because it's the proc's return that's being checked, but I'd recommend using different numbers, so it's clearer at first glance what's happening

saem (approved):
soft suggestion, otherwise LGTM

PR #98, opened 2024-12-24
ci: pin the NimSkull version

zerbina:
## Summary

Pin the NimSkull version used to the most recent working one
(0.1.0-dev.21431), making it the official minimum supported version of
NimSkull to compile the various Phy programs with.

---

## Notes For Reviewers
* this fixes the CI failures caused by the most-recent broken NimSkull release
* [skully](https://github.com/nim-works/phy/pull/70) also needs a pinned version, as it imports compiler modules directly

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -56,10 +56,10 @@ jobs:
           fetch-depth: 0
           filter: tree:0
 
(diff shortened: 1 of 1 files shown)


zerbina:
@saem: The change should be uncontroversial, and I'm merging this eagerly in order to unblock the #97.

PR #99, opened 2024-12-26
lang: use more structured scoping rules

zerbina:
## Summary

By default, expressions now always open a new scope for themselves and
their sub-expression, except for the first operand of `And`/`Or`,
expressions in `Exprs`, and `If` conditions.

## Details

The new scoping rules are much easier to formalize, as a `(Decl a b)`
expressions can now be viewed as `(Let a b trailing)`, where `trailing`
are the following expressions in the enclosing `Exprs`.

The rules arguably also make reasoning about code slightly easier to
the reader, as declarations within some nested sub-expression don't
"escape" outwards anymore.

For the implementation, `scopedExprToIL` is renamed to `exprToIL` and
the former `exprToIL` is renamed to `openExprToIL`.

The change is a preparation for formalizing the language definition.
In the future, some of the previous scoping rules might be restored.

---

## Notes For Reviewers

The simplifications are only really visible when attempting to formalize the rules. The "symbol bleeding" semantics is not easy to model when using conventional outside-to-inside syntax typing judgments, whereas it's easy to judge `(Let x y e)`:
```
C |- y: T   C, x: T |- e: T_2
-----------------------------
    C |- (Let x y e): T_2
```

Operational semantics regarding location lifetimes (and thus cleanup/destruction, once added) also becomes simpler with `Let`, as the LIFO behaviour of locals is encoded directly in the abstract syntax.

The "symbol bleeding" semantics is not impossible to model formally, of course, and it could be worthwhile to attempt doing so in the future.

`Let` is only a theoretical construct at the moment, though it could also be added to the source language.

### DIFF
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -439,7 +439,10 @@ proc fitExprStrict(c; e: sink Expr, typ: SemType): Expr =
 
 proc exprToIL(c; t: InTree, n: NodeIndex, expr, stmts): ExprType
 
-proc exprToIL(c; t: InTree, n: NodeIndex): Expr =
+proc openExprToIL(c; t: InTree, n: NodeIndex): Expr =
+  ## Analyzes the given expression and generates the IL code for it. Symbols
+  ## introduced by the analysis *escape* into the enclosing scope, hence the
+  ## procedure's name.
   result.expr = IrNode(kind: None)
   (result.typ, result.attribs) = exprToIL(c, t, n, result.expr, result.stmts)
   # verify some postconditions:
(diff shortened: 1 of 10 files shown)


passes/source2il.nim
```diff
@@ -439,7 +439,10 @@ proc fitExprStrict(c; e: sink Expr, typ: SemType): Expr =
 
 proc exprToIL(c; t: InTree, n: NodeIndex, expr, stmts): ExprType
 
-proc exprToIL(c; t: InTree, n: NodeIndex): Expr =
+proc openExprToIL(c; t: InTree, n: NodeIndex): Expr =
```
saem, line 442: admittedly this is better than `360NoScopeExprToIL` 😆

saem (approved):

PR #110, opened 2025-01-25
lang: add basic pattern matching (`Match`)

zerbina:
## Summary

Add the `Match` expression, used for pattern-matching over a value. At
the moment, it's only usable for unpacking unions.

## Details

* add the `Match` abstract syntax
* add the static and dynamic semantics of `Match`. The match expression
  must be exhaustive, meaning that a rule must be provided for every
  type in the union
* implement `Match` in `source2il`. It's a straightforward
  implementation of the semantics laid out in `source.nim`. The
  `ekValue` symbol kind is added to be able to represent some immutable
  local entity that behaves like a value (needed for the entity
  introduced by an `As` pattern)
* fix a bug with the meta-language type checker where no error was
  reported for too little arguments in a call

The semantics of `Match` are kept simple so that they're easy to define
and implement. It's intended to be extended gradually over time.

---

## To-Do
- [x] wait for the testable meta-language-based source language definition to be available
- [x] define the semantics for `Match`
- [x] add tests for `Match`


zerbina:
I've changed the syntax from:
```
(Match <expr> (As <ident> <texpr> <expr>)+)
```
to
```
(Match <expr> (Rule (As <ident> <texpr>) <expr>)+)
```

Alternatively, `Case` and `Branch` (the names NimSkull uses for similar entities) could be used for `Match` and `Rule`, respectively, but I personally like the latter more.

The reasoning behind this change is to:
1. make the syntax more easily extensible
2. always use the same shape for `As` (i.e., expression + type)

The idea is that the first element of a `Rule` is the *pattern*. Right now, this can only be an `As` pattern, but in the future this will include things like tuple patterns and literal values (making `Match` a more powerful version of NimSkull's `case`).

saem (approved):

PR #112, opened 2025-01-26
source2il: implement alignment support

zerbina:
## Summary

Implement internal support for locations with an alignment requirement
greater than 1 byte. This is a preparation for the source language
gaining type alignment requirements.

## Details

The alignment for numeric types (int and float) is equal to their size,
while for aggregate types (other than `seq`) it's that of their most
aligned constituent type.

The size of aggregate types is padded to be a multiple of their
alignment. This is so that their size matches the space they would take
up within an array, while still allowing copying `size` bytes from a
pointer to a location of said type to be valid, regardless of where the
location is located. Record fields are placed such that their relative
offset is a multiple of their alignment requirement.

The allocator procedures `alloc` and `realloc` take an additional 
parameter specifying the allocation's required alignment. `grow` and
`preparedAdd` also take an additional parameter, which specifies the
alignment of the dynamic array's element type. The callsites of all
four procedures are updated accordingly.

Finally, the value rendering in `vmexec` is updated to take the new
object layout into account.

---

## Notes For Reviewers
* memory access of numeric types being well-aligned also eases implementation of the `asm.js` translation pass, as `asm.js` doesn't have native support for misaligned memory access

### DIFF
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -311,29 +311,28 @@ proc typeToIL(c; typ: SemType): uint32 =
   of tkTuple:
     let args = mapIt(typ.elems, c.typeToIL(it))
     c.addType Record:
-      c.types.add Node(kind: Immediate, val: size(typ).uint32)
-      # XXX: for the sake of ease of implementation, records use the maximum
-      #      possible alignment
-      c.types.add Node(kind: Immediate, val: 8)
+      c.types.add Node(kind: Immediate, val: paddedSize(typ).uint32)
+      c.types.add Node(kind: Immediate, val: alignment(typ).uint32)
       var off = 0 ## the current field offset
       for i, it in args.pairs:
+        let mask = alignment(typ.elems[i]) - 1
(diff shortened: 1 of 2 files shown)


passes/source2il.nim
```diff
@@ -149,6 +149,12 @@ using
   expr: var IrNode
   stmts: var seq[IrNode]
 
+func align(val: SizeUnit, to: SizeUnit): SizeUnit =
+  ## Rounds `val` to the multiple of `to`, the latter which must be a power of
+  ## two.
```
saem, line 154:
```suggestion
  ## Rounds `val` to the multiple of `to`, the latter must be a power of two.
```

Just playing a bit of word golf to get a one line comment

saem (approved):

passes/source2il.nim
zerbina: Yea, the short-one-word-comments are a bit awkward.

PR #113, opened 2025-01-30
langdefs: use `case` expression for function bodies

zerbina:
## Summary

Instead of a list of assignments of the form `<f> <pattern>* = <body>`,
the body of a meta-function must now be a `case` expression.

## Details

A function such as:
```
function f, x -> x:
  f(z) = 1
  f(1) = 2
```
is now written as:
```
function f, x -> x:
  case _
  of z: 1
  of 1: 2 
```

The two reasons for this change are:

1. there's no point in repeatedly specifying the name of a function:
   it's the only valid identifier that may appear in before the
   parenthesis. It's presence only introduces visual noise and makes
   it harder to rename functions

2. the assignment based syntax is a bit trickier to parse, and it's
   arguably not very intuitive. For people already familiar with
   NimSkull and its syntax, what the new `case` expression syntax does
   is likely easier to understand

Using a `case` expression also makes `else` available to be used for
representing the fallthrough case (the new parser already supports
this).

Two things that might be slightly confusing are that:
1. contrary to the NimSkull semantics, the `of` branches in the
   DSL are ordered
2. in the DSL, the body of the `of` branches also affects whether the
  branch is actually "picked"

The macro's parsing logic is updated to handle the new syntax, and the
source language specification is changed to make use of it.

---

## Notes For Reviewers

Personally, I think the new syntax is better and more intuitive (despite the differences in semantics compared to NimSkull), but I'm also okay with keeping the previous syntax (and thus not merging this PR).

### DIFF
--- a/languages/source.nim
+++ b/languages/source.nim
@@ -110,17 +110,18 @@ const lang* = language:
 
   function desugar, expr -> e:
     # FIXME: the sub-expressions need to be desugared too!
-    desugar And(expr_1, expr_2) =
+    case _
+    of And(expr_1, expr_2):
       If(expr_1, expr_2, "false")
-    desugar Or(expr_1, expr_2) =
+    of Or(expr_1, expr_2):
       If(expr_1, "true", expr_2)
-    desugar Decl(x_1, expr_1) =
+    of Decl(x_1, expr_1):
       Let(x_1, expr_1, TupleCons())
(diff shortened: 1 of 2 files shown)


saem (approved):
At first I thought we could have bare `of` branches, but then I realized that's probably a grammar issue, I think the `case _` is a small "price" to pay for something that writes about as elegantly as elixir or ocaml.

PR #115, opened 2025-02-05
spec: clarify `readFile` semantics

zerbina:
## Summary

Express the semantics of `readFile` using the currently available
meta-language features.

## Details

The previous idea boiled down to treating the result of `readFile` as
fully random (within the bounds of the value's inhabited type). This is
wrong, as it *is* possible to precisely know the value, where afforded
by the context.

The dynamic execution context is extended with a function that maps a
file path and "time" to a UTF-8 character sequence, which is used to
provide the result of a file system read access. The purpose of the
time value is to model the fact that successive reads with the same
path may yield different results, without the program changing anything
itself. Without the time parameter, the definition - due to the meta-
language's side-effect free nature - would define `readFile` as always
yielding the same value for successive calls with the same path.

---

## Notes For Reviewers
* a small improvement split out from the in-progress definition rework

### DIFF
--- a/languages/source.nim
+++ b/languages/source.nim
@@ -733,12 +733,6 @@ const lang* = language:
     rule "E-builtin-concat":
       conclusion Call("concat", `array`(*val_1), val_2), `array`(...val_1, val_2)
 
-    rule "E-builtin-readFile":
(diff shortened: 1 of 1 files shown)


saem (approved):
that's really clever, I like it

PR #118, opened 2025-03-03
phy: use structured output formatting

zerbina:
## Summary

* show program result value and its type as S-expression (easier to
  process)
* support positional output specs (e.g., `output.0: ...`) in test
  specification
* separate program output from `phy` output when in runner mode

## Details

* return an S-expression-based representation of the result (value and
  type) from `vmexec.run`. It matches how the value would be represented
  in the source language definition
* when in runner mode, separate the program output (i.e., text written
  to the output and error stream) from the value + type, so that both
  count as separate fragments for the tester and can be compared
  independently
* support positional output specs in test specification to make it
  possible to provide the expected string for leading fragments
  directly, without the need to use `.expected` files
* update the `expr` tests' `output` specifications

The output is now much easier to parse programmatically.

---

## Notes For Reviewers
* a preparation for the source language specification tests

### DIFF
--- a/phy/vmexec.nim
+++ b/phy/vmexec.nim
@@ -3,11 +3,12 @@
 
 import
   std/[
-    options,
-    strutils
+    options
+  ],
+  experimental/[
+    sexp
   ],
   phy/[
-    type_rendering,
     types
   ],
   vm/[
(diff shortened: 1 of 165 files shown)


phy/vmexec.nim
```diff
@@ -27,94 +28,79 @@ type
...
 proc readInt(p: HostPointer, size: range[1..8]): int64 =
   copyMem(addr result, p, size)
 
 proc readFloat64(p: HostPointer): float64 =
   copyMem(addr result, p, 8)
 
-proc primToString(v: Value, typ: SemType): string =
-  ## Renders the primitive value `v` of `typ` type to a string.
+proc primToSexp(v: Value, typ: SemType): SexpNode =
+  ## Renders the primitive value `v` of `typ` type to an S-expression.
```
saem, line 44:
```suggestion
  ## Renders the primitive value `v` of type `typ` to an S-expression.
```

tools/tester.nim
```diff
@@ -414,14 +416,33 @@ proc runTest(desc: RunnerDesc, file: string): bool =
...
+            OutputSpec(fromFile: false,
+                       output: strip(evt.value, leading=true, trailing=false))
+          if evt.key == "output":
+            if usesPositional:
+              echo "cannot mix positional and non-positional 'output'"
+              return false
+            spec.expected.add output
+          else:
+            let i = find(evt.key, '.')
+            if i == -1:
```
saem, line 434: should this also check to see if '.' doesn't appear as the last character, as that would also be illformed?

saem (approved):
Two minor things you might wish to consider, but LGTM

tools/tester.nim
zerbina: That's already taken care of indirectly, via the following `parseInt` call. If the string is empty or contains non-digit characters, an exception will be raised.

saem: Yup, I did see that, I should have mentioned it, I just meant _if_ you wanted to opt for a more specific error message, but all this is a soft suggestion.

I should add the benefit of the exception is that it's still precise, even if a little indirect.

zerbina: I'll add dedicated handling for the missing number case.

PR #119, opened 2025-03-04
interpreter: represent all numbers as rationals

zerbina:
## Summary

* implement proper support for rational numbers in the meta-language
  interpreter
* fix non-integer `FloatVal` not being considered values in the source
  language definition
* significantly improve compile times of `language` and `term` macro
  invocations

## Details

* add the `Rational` type, implemented as a numerator/denominator pair
* implement the basic arithmetic and comparison operators for `Rational`
* store all numbers as `Rational`s in `Node`. All usage sites of `Node`
  expecting numbers stored as strings in `.sym` are updated
* update `z` pattern matching in the interpreter to only match integer
  numbers
* remove `newLit` usage from `language` and `term`; `newLit` cannot be
  used anymore, as the fields of `Rational` (stored in `Node`) are not
  exported. The actual processing is factored into normal procedures
  taking a `NimNode` as input, with the macros then emitting a call to
  the respective procedures, passing the quoted macro argument along
* use `rat` (instead of type variables) as the parameter type for some
  built-in functions, to improve type checking precision and type
  inference

The `newLit` usage majorly contributed to compile times, because:
1. the `newLit` call itself has to process a large amount of data
2. the (potentially huge) construction expressions emitted by the
   macros need to be sem-checked again *and* - if appearing in a
   `const` context - compiled for and evaluated with the VM

Removing the `newLit` usage therefore significantly decreases compile
times. 

---

## To-Do
- [x] narrow down the built-in numeric operations' types
- [x] add tests (requires some unit test infrastructure first)


zerbina:
To not lose information for rationals that don't have a (finite) string representation, `Node` must store numbers as `Rational`s directly. This is a bit more effort to implement, but the upside is that it'll get rid of the repeated parse/stringify operations in the interpreter.

zerbina:
It would have also been possible to render fractions with not finite decimal number representation as `a/b`, but this would have slightly complicated both parsing/rendering, while not addressing the repeated parsing/stringification when operating on the numbers in the interpreter.

zerbina:
On my machine, the compile time savings are at around 1 second, which is quite good.

zerbina:
I've fixed an edge case with signed 128-bit integer division and `reduce` where they got stuck in an infinite recursion (followed by a crash due to a stack overflow).

@saem: The PR is ready for review now, but please don't merge it yet. I want to first make sure that everything still works after #127 has been merged.

zerbina:
Okay, back to a draft it goes.

Rational numbers (for usage as the numerical value representation in the interpreter) needs arbitrary precision to be practically useful. My previous plan was to first merge rational number support and then implement and use some simple bignum library in post, but I'm almost done with implementing the bignum module already, so there's no point in merging a (mostly) useless rational number library first.

zerbina:
Why not use the generic `Rational` type from `std/rationals`, which is more feature complete? Because the type parameter is constrained such that it to only admits built-in integer types.

The standard library `Rational` type *could* be changed to allow all types as the generic parameter, however. Constraining the type parameter with a concept would be better, of course, but I don't think concepts are stable enough for that.

spec/rationals.nim
```diff
@@ -0,0 +1,193 @@
...
+
+type Rational* = object
+  ## A rational number represented as a fraction of two arbitrary-precision
+  ## integers.
+  num, den: Bignum
+
+const
+  One = 1'n
+
+proc reduced(r: sink Rational): Rational =
```
saem, line 16: Any reason behind the preference for the past tense?

`myRational.reduced` vs `myRational.reduce`... they're not super different, I don't see the need to change it, just curious

saem (approved):
LGTM, I'm going to merge it, the question I left in the comment can be discussed async.

spec/rationals.nim
zerbina: > Any reason behind the preference for the past tense?

Yep, it's because I'm using the naming convention from Nim/NimSkull, where routines that modify their inputs in-place (via a `var` parameter) use the imperative, whereas routines that don't use past tense.

`reduced` could also reduce the rational in-place (in which case the procedure would be named `reduce`), but then all callsites would need to use either temporaries or the result variable, which is longer, and I also have a slight preference towards a more functional style.

PR #120, opened 2025-03-04
test the reference implementation

zerbina:
## Summary

Also run the `expr` tests using the source language reference
implementation `languages/source.nim`, to make sure that the
reference/specification implements/describes the desired semantics.

## Details

As a temporary solution, a standalone test orchestrator is used that
scans the `expr` directory for test files. For each test file, the
spec is extracted and (the relevant bits of it) parsed, with the test
content then parsed into a representation understood by the meta-
language interpreter.

The processing pipeline is: desugar -> type-check -> evaluate. Stages
are enabled/disabled based on the test specification. Failing tests
don't result in an orchestrator error, so that they can be fixed
incrementally; the list of known issues is hard-coded and needs to be
updated manually once more tests start to succeed.

There are many issues with the reference implementation as is, and
therefore not all tests can be run successfully at the moment.

A few small changes/additions are made to `source`, so that at least
some tests work:
* the `cstep` relation is added to `source`, which is the transitive
  closure of `step`. It's used for fully reducing (i.e., evaluating)
  an expression
* an infinite recursion with the `<:=` relation is fixed
* a catch-call branch is added to `desugar` so that the function
  doesn't fail for unhandled input

Finally, a dangling reference caused by a parameter aliasing violation
with the `mitems`'s iterator usage in `interpretRelation` is fixed.


zerbina:
The current plan going forward is as follows:
1. incrementally fix all reference implementation issues
2. once all issues are resolved (and thus all tests succeed), replace the orchestrator with a proper test runner
3. make writing tests easier by grouping multiple related tests into a single test file

Work on new language features can continue in parallel. All tests not included in the `issues` array are required to succeed, which is going to prevent reference implementation regressions during this time.

saem (approved):

zerbina:
#121 introduced a parameter aliasing violation leading to the previous CI failure. I've fixed it and updated the PR message accordingly.

PR #122, opened 2025-03-06
lang: add explicit type widening (`As`)

zerbina:
## Summary

Add the `(As x typ)` expression to the source language, which is used
to explicitly widen the type of a value. `x` having a type that is
equal to `typ` is not an error.

## Details

`As` takes an expression and a type, fitting the former to the latter
where possible, reporting an error if not. At the moment, it's main
use is constructing values of `UnionTy` type, but in the future, it'll
also be useful for guiding type inference.

### Implementation

* lower `As` directly into a type conversion through `fitExpr` 
* fix error correction for fitting errors producing illformed `Expr`s,
  by keeping or discarding the trailing expression depending on the
  target type

Tests are added for `As`, and existing tests for union types now use
`As` instead of helper procedure for creating union values, greatly
simplifying the affected tests.

---

## Notes For Reviewers
* needs to wait for #120 to be merged first
* the feature is introduced now in order to make union-using tests (such as the ones for #110) easier to write


passes/source2il.nim
```diff
@@ -1274,6 +1278,13 @@ proc exprToIL(c; t: InTree, n: NodeIndex, expr, stmts): ExprType =
         newCall(at, @[c.inlineLvalue(arr, stmts), c.capture(idx, stmts)]))
     else:
       result = errorType() + arr.attribs
+  of SourceKind.As:
+    let
+      (a, b) = t.pair(n)
+      e = c.exprToIL(t, a)
+      typ = c.expectNot(c.evalType(t, b), tkVoid)
+    expr = inline(c.fitExpr(e, typ), stmts)
+    result = typ + {} # lvalue-ness and mutability are discarded
```
saem, line 1287: I'm still learning to read the source, but it seems that lvalue-ness and mutability dropping isn't there, I'm guessing it can't be represented?

I guess the drop is required right now because if we're actually widening the bits then a new location would be required?

saem (approved):
Just some questions, otherwise LGTM

passes/source2il.nim
zerbina: > I'm still learning to read the source, but it seems that lvalue-ness and mutability dropping isn't there, I'm guessing it can't be represented?

It's there, albeit not easy to spot. `exprToIL` (the enclosing procedure) returns an `ExprType`, which is a type + qualifiers (`ExprFlags`, currently a bunch of flags), where the qualifiers track both mutability and lvalue-ness.

If the qualifiers were to be kept, the result assignment would look like this:
```nim
result = typ + e.attribs # e.attribs are the source expression's qualifiers 
```

> I guess the drop is required right now because if we're actually widening the bits then a new location would be required?

Yep, that's the idea. If no widening takes place, then no new location is needed, so the lvalue-ness and mutability *could* be kept, but I decided against it because:
* to the human reader, a large context window is potentially required to know what a specific `As` does (the same problem exists in NimSkull with lvalue conversions)
* the lvalue-ness/mutability changing depending on the target type could become an issue once/if per-procedure type inference is introduced

This doesn't have to (nor should it have to) be the final decision, however. If it turns out that the current rules are problematic (for whatever reason), we should alter them.

zerbina:
@saem: The dynamic semantics for `As` are now correct.

If you want to see what the difference in evaluation would be were `As` to preserve l-value-ness, you can do the following:
1. remove the `As(hole, texpr)` rule from `F`
2. change the `E-as` rule to `axiom "E-as", As(e_1, texpr), e_1`

Then run the following commands in the `queryshell`:
```sh
> (import "languages/source") # load the source language
> (default DC) (set dc)       # get a default dynamic context
> (term (Exprs
          (Decl (Ident "x") (Seq (IntTy) (IntVal 1)))
          (At
            (As (Ident "x") (SeqTy (IntTy)))
            (Exprs
              (Asgn (Ident "x") (Seq (IntTy) (IntVal 2)))
              (IntVal 0)))))   # this is the expression from `t12_as_copies.test`
> (apply desugar) (set t) # desugar and save the term to a variable, for easier subsequent access
> (apply step $dc $t)     # perform a single step on `t` using context `dc`
> (apply step)            # perform a single step on the previous step's output. Repeat...
```

PR #124, opened 2025-03-08
spec: add missing boolean rules

zerbina:
## Summary

* add reduction rules for `Ident("true")` and `Ident("false")`
* add reduction and type deduction rules for `not`


zerbina:
It's also possible for `Ident("true")` and `Ident("false")` to be considered values, in which case `True` and `False` would become `def`s rather than meta-language values / type constructors.

Intuitively, I would say that the current approach is better in the short term, though it might have to be revisited in the longer term.

saem (approved):

PR #127, opened 2025-03-10
spec: fix dynamic semantics for lvalues

zerbina:
## Summary

* make the reduction of lvalue expressions model the intended dynamic
  semantics
* fix two small bugs with the meta-language semantic analyzer

## Details

There were multiple interacting issues with the previous reduction
rules.

#### `loc` As Value

`loc(z)` was considered a value (i.e., irreducible expression) even
though it technically isn't one. The idea was to have `loc`s be handled
by the reduction rules for their enclosing syntax (by `copy`ing them,
or not), but this introduces a lot of complexity and also wasn't
considered everywhere.

The `loc(z)` syntax is no longer considered a value, and the `tuple`
syntax used to represent reduced tuple values plus the `E-tuple-cons`
rule become obsolete. `TupleCons(*val)` is now considered a value,
which also addresses reduction getting stuck on `TupleCons()` (unit).

#### `loc` Reduction

A first-class location (i.e., `loc(...)`) cannot be reduced (by
replacing it with the stored value) immediately in all contexts.
Exceptions need to be made for:
* the left-hand side of assignments (e.g., `Asgn(loc(...), ...)`)
* the root of projections (`FieldAccess` and `At`)

The intended semantics are for projections are "full lvalue expressions
are reduced only once all non-lvalue operands were reduced", which
`E-tuple-loc-access` and `E-at-loc` didn't model properly, as they'd
reduce the loc in a not-fully-evaluated expression such as
`At(At(loc(...), 1), Call(+, 1, 2))`.

The new `F` context is introduced to select the next "full" expression,
that is, an expression in a position where lvalues must be eagerly
reduced to values. It's used by the new `E-load` and `E-load-top`
rules.

#### `Exprs` Folding

`Exprs` may act as an lvalue expression, by having a trailing lvalue
expression and itself appearing in an lvalue position. So that
`Exprs` are folded without trailing sub-expression being reduced first,
`Exprs(E, *e)` is changed to `Exprs(E, +e)` (so that expressions in 
single-item `Exprs` are never evaluated), and `E-exprs-fold` is changed
to always fold all single-item `Exprs`.

### Meta-language Bugs

* fix types erroneously being allowed in expression positions, leading
  to the interpreter crashing on encountering them
* fix `hole` not being usable in r-pattern-matcher's patterns. Only the
  type (previously `all`) is considered by the type checker's pattern
  matching, and `all` only matches with type `all`. `void` (which is
  compatible fits all other types) is the correct type for `hole`


zerbina:
The fix is quite fundamental and is going to unblock a good amount of other work. For example, the `As` expression, array types in the source language, dynamic tuple indexing, and record type support, were all affected by the incorrect lvalue semantics.

saem (approved):

PR #130, opened 2025-03-14
use arbitrary-precision integers in the meta language 

zerbina:
## Summary

* add a native `Bignum` implementation (`bignums.nim`)
* use `Bignum` for all arithmetic/comparison operations in the
  interpreter

## Details

A NimSkull-native `Bignum` implementation is used because:
* porting it to the source language in the future will be easier / can
  be automated
* the repository stays self-contained

Two explicit goals the implementation were that it's fully usable at
compile-time (in the VM) and that zero-initialized `Bignum`s are valid.
The implementation tries to strike a balance between complexity and
efficiency, usually preferring a simpler over a more efficient
implementation.

Using two's complement representation gets around having an extra sign
flag, which would increase the size of `Bignum`s by 8 byte (on a 64-bit
target).

Tests based on the `int128` tests are added for `Bignum`, with new
tests added for the various overflow/carry/extension scenarios.

---

## Notes For Reviewers
* a prerequisite for #119
* the `int128` module - while now unused - is kept because it'll likely be needed again in the future

### DIFF
--- a/tests/unittest/tbignums.nim
+++ b/tests/unittest/tbignums.nim
@@ -1,0 +1,283 @@
+discard """
+  description: "Tests for the 128-bit integer implementation"
+"""
+
+import spec/bignums {.all.}
+
+const Full = high(uint64) # all 64 bits are set
+
+# don't use 'n for early tests, as it relies on parsing, which itself relies
+# on working addition and multiplication
+proc bn(x: SomeInteger): Bignum =
+  bignum(x)
+
+block add_with_carry:
+  # test the internal add-with-carry implementation
+  doAssert adc(1, 1, false) == (2'u64, false)
+  doAssert adc(Full, 1, false) == (0'u64, true)
+  doAssert adc(Full, 0, true) == (0'u64, true)
+  doAssert adc(Full, 1, true) == (1'u64, true)
+  doAssert adc(Full, Full, false) == (Full shl 1, true)
+  doAssert adc(Full, Full, true)  == (Full, true)
+
+block negation:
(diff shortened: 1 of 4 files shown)


tests/unittest/tbignums.nim
```diff
@@ -0,0 +1,283 @@
...
+
+  doAssert sum == 111111111111111111111111111111111111111'n
+
+  # test subtraction:
+  for it in d.items:
+    sum = sum - it
+
+  doAssert sum == Zero
+
+# test comparison, multiplication, and division with positive nubers:
```
saem, line 161:
```suggestion
# test comparison, multiplication, and division with positive numbers:
```

tests/unittest/tbignums.nim
```diff
@@ -0,0 +1,283 @@
...
+for i, a in d.pairs:
+  for j, b in d.pairs:
+    doAssert(cmp(a, b) == cmp(i, j))
+    if i + j < d.len:
+      doAssert a * b == d[i + j]
+    if i - j >= 0:
+      doAssert a div b == d[i - j]
+
+# test comparison, multiplication, and division with negative nubers:
```
saem, line 170:
```suggestion
# test comparison, multiplication, and division with negative numbers:
```

saem (approved):
I found two minor typos, but otherwise looks good.

I was expecting that nim's `bignum` would use `array[4, uint64]` or the like. Either that or save on the pointer, capacity, and size fields by consolidating them and then shrinking things further. But that's all out of scope, imo.

PR #131, opened 2025-03-17
spec: define floating-point operational semantics

zerbina:
## Summary

Define the operational semantics of the current source language float
operations according to the 2008 revision of the IEEE-754 standard,
making their behaviour well-defined.

## Details

Why the 2008 revision? Because:
* it uses the contemporary terminology (e.g., subnormals instead of
  denormals)
* it includes 16-bit floats, for whom support is planned in the future

Some simplifications are made, to keep the specification simpler for
the moment:
* there's no error handling beyond NaN propagation
* NaNs have no sign nor payload
* rounding always uses the default to-nearest, tie-to-even mode

Some basic tests are added, covering some rounding cases and most cases
involving non-finite values (e.g., `inf + nan`). They need to be
extended once writing many similar tests is made easier.

To get around stack overflow crashes in the interpreter due to too deep
recursion, the `digit2` function tries to right-shift by more than 1 per
iteration. 

---

## To-Do
* [x] merge the `main` branch once the necessary additions have landed
* [x] extend the floating-point operation tests
* [x] improve the documentation


zerbina:
The initial definition was largely based off of the float specification from Coq's standard library. I've changed the rounding to use rational numbers instead of bit-shifting, the former which is much simpler (and also faster).

The way addition and subtraction are implemented is still a problem, as - depending on the exponents of the operands - it currently needs numbers with very large magnitudes (way beyond what `Int128` supports). It's possible to change the implementation to stay well inside the supported `Int128` range, but I want to wait until the current implementation works before changing it.

zerbina:
Interesting, it looks like either `-d:nimPreviewFloatRoundtrip` or the `tree_parser` changes result in `skully` miscompiling `phy`. I'll investigate.

zerbina:
For the tests, my plan is to support providing multiple test cases per test file, which will make it much easier to inspect and modify them, since related test cases (e.g., for float addition) are all in one place.

For the interpreter stack overflows, the plan is to use proper continuation-passing style for `interpret` and friends, making use of NimSkull's new `.tailcall` calling convention.

saem (approved):

PR #132, opened 2025-03-20
lang: allow non-void expressions as bodies

zerbina:
## Summary

Allow any non-void expression as procedure's body, as long as its type
fits the return type. This reduces `Return` noise and allows for more
functional-style procedure declarations.

## Details

* update the language definition
* implement the feature in `source2il`, by passing the body expression
  to `fitExpr` and wrapping non-void expression in a `Return` expression

---

## Notes For Reviewers
* a small out-of-band improvement to the language


saem (approved):
That came together really nicely, the consistency of handling things really makes composition smooth

PR #134, opened 2025-03-21
spec: properly implement `desugar`

zerbina:
## Summary

Implement/define the desugaring logic properly, so that it works as
originally intended.

## Details

* the desugar transformation needs to be applied recursively to all
  transitive sub-expressions, not just to the top-level expression
* in order to implement the bubbling-up of `Exprs`, two passes are used:
  one for sugar expansion + propagation, and one for the `Decl`
  transformation
* the `update` function is introduced and used so that the actual
  transformation logic is separate from the traversal logic

Tests using nested `Or`, `And`, and `Decl` - and not using other not-
yet-working features - now succeed.


zerbina:
The PR is ready for review, but merging it needs to wait until the other open specification PRs are merged, so that there are no explicit or implicit merge conflicts.

languages/source.nim
```diff
@@ -100,23 +100,76 @@ const lang* = language:
     FieldAccess(le, IntVal(n))
     At(le, IntVal(n))
 
-  func desugar(a: expr) -> e =
-    # FIXME: the sub-expressions need to be desugared too!
+  func update(a: expr, xfrm: e -> e) -> expr =
```
saem, line 103: This is really nice, it will help avoid a lot of bugs (missed conditions in otherwise the same traversal) and promote consistency in designing rules language features.

saem (approved):

PR #135, opened 2025-03-22
lang: add dynamic tuple indexing

zerbina:
## Summary

Extend `At` to also allow tuple values as the first operand. If the
tuple is homogeneous, the expression has the type of the tuple element,
otherwise it's a union of all possible element types.

## Details

* add a typing rule for `At` with tuples
* add reduction rules for `At` and `With` covering tuples
* implement `At` support for tuple values in `source2il`
* fix a bug with `source2il.align` that resulted in the wrong value
  being returned (the mask was computed incorrectly)

`At` with a tuple operand never yields a mutable lvalue, even if the
input is one, as it'll make the future initial implementation of
pattern matching easier, but this decision is subject to change.

---

## Notes For Reviewers
* depends on #122 to work

### DIFF
--- a/languages/source.nim
+++ b/languages/source.nim
@@ -296,7 +296,7 @@ const lang* = language:
     # of types
     case list
     of [typ_1, *typ_2]:
-      if typ_1 in typ_2:
+      if contains(typ_2, typ_1):
         # it's a duplicate; skip
         unionOfAux(typ_2, accum)
       else:
--- a/languages/source.nim
+++ b/languages/source.nim
@@ -288,6 +288,28 @@ const lang* = language:
     of type(typ): true
     of typ:       false
 
+  func unionOfAux(list: *typ, accum: typ) -> typ =
+    # `TupleTy` is used as the accumulator, because it can hold any number
+    # of types
+    case list
+    of [typ_1, *typ_2]:
+      if typ_1 in typ_2:
+        # it's a duplicate; skip
+        unionOfAux(typ_2, accum)
+      else:
+        case accum
+        of TupleTy(*typ_3):
+          unionOfAux(typ_2, TupleTy(...typ_3, typ_1))
+    else:
+      accum
+
+  func unionOf(list: *typ) -> typ =
+    case unionOfAux(list, TupleTy())
+    of TupleTy(typ_1, +typ_2):
+      UnionTy(typ_1, ...typ_2)
+    of TupleTy(typ_1):
+      typ_1
+
   inductive types(inp C, inp e, out typ):
     axiom "S-int",   C, IntVal(n), IntTy()
     axiom "S-float", C, FloatVal(r), FloatTy()
@@ -349,6 +371,11 @@ const lang* = language:
       premise mtypes(C_1, e_2, IntTy())
       where SeqTy(typ_2), typ_1
       conclusion C_1, At(e_1, e_2), mut(typ_2)
+
+    rule "S-at-tuple":
+      premise mtypes(C_1, e_1, TupleTy(+typ_2))
+      premise mtypes(C_1, e_2, IntTy())
+      conclusion C_1, At(e_1, e_2), unionOf(typ_2)
 
     rule "S-asgn":
       premise types(C_1, e_1, mut(typ_1))
@@ -763,13 +790,27 @@ const lang* = language:
       let val_2 = val_1[n_1]
       conclusion At(array(*val_1), IntVal(n_1)), val_2
 
+    rule "E-at-tuple":
+      condition 0 <= n_1
+      condition n_1 < len(val_1)
+      let val_2 = val_1[n_1]
+      conclusion At(TupleCons(*val_1), IntVal(n_1)), val_2
+
     rule "E-at-out-of-bounds":
       condition n_1 < 0 or len(val_1) <= n_1
       conclusion At(array(*val_1), IntVal(n_1)), Unreachable()
 
+    rule "E-at-tuple-out-of-bounds":
+      condition n_1 < 0 or len(val_1) <= n_1
+      conclusion At(TupleCons(*val_1), IntVal(n_1)), Unreachable()
+
     rule "E-with-out-of-bounds":
       condition n_1 < 0 or len(val_1) <= n_1
       conclusion With(array(*val_1), n_1, val_2), Unreachable()
+
+    rule "E-with-tuple-out-of-bounds":
+      condition n_1 < 0 or len(val_1) <= n_1
+      conclusion With(TupleCons(*val_1), n_1, val_2), Unreachable()
 
     axiom "E-not-false", Call(Ident("not"), False), True
     axiom "E-not-true",  Call(Ident("not"), True),  False
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -151,7 +151,7 @@ using
 
 func align(val: SizeUnit, to: SizeUnit): SizeUnit =
   ## Rounds `val` to the multiple of `to`, the latter must be a power of two.
-  let mask = val - 1
+  let mask = to - 1
   (val + mask) and not mask
 
 template `+`(t: SemType, a: set[ExprFlag]): ExprType =
--- a/passes/source2il.nim
+++ b/passes/source2il.nim
@@ -583,24 +583,71 @@ proc genTypeBoundOp(c; op: TypeAttachedOp, typ: SemType): uint32 =
     else:
       unreachable() # no copy procedure needed
   of opAt:
-    # at(s: seq(T), index: int) -> pointer
-    procTy = SemType(kind: tkProc, elems: @[pointerType, typ, prim(tkInt)])
-    params.add 0
-    params.add 1
-    locals.add typ
-    locals.add prim(tkInt)
+    case typ.kind
+    of tkSeq:
+      # at(s: seq(T), index: int) -> pointer
+      procTy = SemType(kind: tkProc, elems: @[pointerType, typ, prim(tkInt)])
+      params.add 0
+      params.add 1
+      locals.add typ
+      locals.add prim(tkInt)
 
-    let
-      s   = newLocal(0)
-      idx = newLocal(1)
+      let
+        s   = newLocal(0)
+        idx = newLocal(1)
 
-    body = IrNode(kind: Stmts)
-    body.add newIf(
-      c.newIntOp(Le, newIntVal(0), idx),
-      newIf(
-        c.newIntOp(Lt, idx, newFieldExpr(s, 0)),
-        newReturn(newAddr(c.genSeqAccess(s, idx, typ)))))
-    body.add newUnreachable()
+      body = IrNode(kind: Stmts)
+      body.add newIf(
+        c.newIntOp(Le, newIntVal(0), idx),
+        newIf(
+          c.newIntOp(Lt, idx, newFieldExpr(s, 0)),
+          newReturn(newAddr(c.genSeqAccess(s, idx, typ)))))
+      body.add newUnreachable()
+    of tkTuple:
+      # at(s: tuple, index: int, tmp: pointer) -> pointer
+      let (ret, folded) = unionOf(typ.elems)
+      procTy = SemType(kind: tkProc,
+                       elems: @[pointerType, typ, prim(tkInt), pointerType])
+      params.add 0
+      params.add 1
+      params.add 2
+      locals.add typ
+      locals.add prim(tkInt)
+      locals.add pointerType
+      locals.add prim(tkBool) # temporary
+      let
+        tup = newLocal(0)
+        idx = newLocal(1)
+        res = newDeref(c.typeToIL(ret), newLocal(2))
+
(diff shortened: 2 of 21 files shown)


zerbina:
After having pushed the branch, I've realized that there's a better static semantics possible than the one used now.

Instead of being typed as the union type created from the set of all tuple element types, `At` should be typed as the *common type* of the tuple elements, but with introduction/merging of `UnionTy`, such that:
* the common type between `UnionTy(IntTy, FloatTy)` and `UnitTy` is `UnionTy(IntTy, FloatTy, UnitTy)`
* the common type between `UnionTy(IntTy, FloatTy)` and `UnionTy(FloatTy, UnitTy)` is `UnionTy(IntTy, FloatTy, UnitTy)`
* the common type between `FloatTy` and `IntTy` is `UnionTy(FloatTy, IntTy)`

A consequence of using the above semantics is that there would be no nested union in `t04_tuple_at_10.test` and `t04_tuple_at_11.test`.

zerbina:
Separate from the above, there's something more general to think about.

When viewing types as nothing more than sets of values, arrays are nothing more than heterogeneous tuples, with an array constructor being the type equivalent of exponentiation:
1. `type ^ 0 = unit`
2. `type ^ 1 = type`
3. `type ^ 2 = type * type` (i.e., a binary tuple / pair)
4. `type ^ n = type * type * type * ...` (i.e., an n-ary tuple)

This perspective is quite neat, because applying it to the source language would mean that there needs to be no standalone array type (dedicated constructors for homogeneous tuples and tuple types will still be needed).

The problem: the current source language doesn't view type as just sets of values. Examples:
1. `(UnionTy (IntTy) (UnionTy (IntTy) (FloatTy)))` != `(UnionTy (IntTy) (FloatTy))`
2. `(TupleTy (IntTy))` != `(IntTy)`

In both cases, both types represent the exact same set of values, but the types are considered not equal. Number 1 was a deliberate choice (to allow for better composition, especially once type inference, type variables, and type parameters are added), number 2 less so.

Nothing of the above is a blocker for this PR, but it does hint at the current type theory not being formal / principled enough.

zerbina:
> After having pushed the branch, I've realized that there's a better static semantics possible than the one used now.

Hm, upon trying us use the mentioned semantics, I realized that it's only feasible when nested unions are allowed. While not impossible for `source2il` to implement these semantics, it - given the current `source2il` union implementation - would potentially require a lot of repacking at run-time, which I think is not right.

I'm aware that I advocated for the `UnionTy` constructor not flattening the resulting union in the past, but I think this design choice needs to be revisited. My recommendation is to wait until there's at least one kind of type in the language that can have a subtype relationship with a non-union, non-void type, so that the `a + b` case where `a` is a sub- or super-type of `b` can be considered in the decision-making process.

zerbina:
There was a subtle issue with the `source2il` implementation of `At` regarding the evaluation order, which I've now fixed. Consider:
```
(Exprs
  (Decl x (TupleCons (TupleCons 1 2)))
  (At
    (At x 3)
    (If true (Return 10) 1)))
```
The inner tuple access is an out-of-bounds access (the tuple only has two elements). However, according to the language definition, this snippets evaluates to 10.

This is because `At` uses "layered evaluation", that is, the non-lvalue operands of the total lvalue expression (in this case `(At (At ... ...) ...)`) are evaluated first, and afterwards the lvalue expression is collapsed. Since the outer `At`'s index expression reduces to a `Return`, collapsing the lvalue expression (and thus the index check) never takes place.

Somewhat counterintuitively, inlining the tuple construction into the the inner `At` would make the snippet reduce to an `Unreachable` (I've added tests for these cases).

`source2il` previously produced code that always used eager evaluation, which is now fixed, albeit in a rather hack-y fashion. I'm not too worried about the hack-y nature of the fix, however, as it's easy to do properly with a new IL that encodes the expected layered evaluation.

The evaluation order of `At` needs to be revisited the future, but for now, I believe the semantics to be fine.

saem:
> Separate from the above, there's something more general to think about.
> 
> When viewing types as nothing more than sets of values, arrays are nothing more than heterogeneous tuples, with an array constructor being the type equivalent of exponentiation:

I'd thought about this before when I was thinking about reworking numskull's type system, but I can't find the relevant gist which enumerated all the types I'd thought through, where my thinking left off was:

There are a few major points around "vector types" (array, tuple, etc), as opposed to "scalar" (bool, int, char, etc), that "vector types" have:

- mixture: heterogeneous or homogeneous elements (i.e. `array` vs `tuple`)
- indexing:
  - computed: homogeneous types have fixed width simplified indexing `offset? + num * element_width` minimal static info, while
  - lookup: heterogeneous types have variable width map based indexing `offset? + index_lookup(num)` K/V static info
- ordering: index and fields are ordered for arrays, sequences, tuples, while not for objects
- sizing: are growable or statically sized (i.e. `seq` vs `array`)

The above gives us the following:

|name|mixture|indexing|ordering|sizing|
|------|-------|---------|--------|------|
|array|homogeneous|computed|ordered|fixed|
|enum array|homogeneous|lookup|ordered|fixed|
|sequence|homogeneous|computed|ordered|variable|
|tuple|heterogeneous|lookup|ordered|fixed|
|object|heterogeneous|lookup|unordered|fixed|
|1???|heterogeneous|lookup|unordered|variable|
|2???|heterogeneous|lookup|ordered|variable|

Note 1: I'm only using `sequence` here because of familiarity with the term, but I'm not convinced it's the best term for the kind itself.

Of course these aren't all the permutations, but that's because a bunch of the combinations aren't that useful in reality, a homogeneous, fixed width, unordered, lookup structure is not meaningfully better than an ordered one (enum array). Who knows, maybe I'll be proven wrong in the future. `1???` and `2???` feel like memristors, postulated about for some time, but we lack the technology/understanding to realize them, I have my suspicions as to where they're hiding though.

I'm inclined to call all these types "vector" types (kinds?) as each field can be considered a dimension, sometimes all the dimensions can be the same (e.g. spatial dimensions in 3D), sometimes they can vary (e.g. 4D 3 spatial dimensions + time).

I'm sharing the above to show how our thinking seems to be in the same neighbourhood, but I was approaching it more from a property/taxonomy point of view, and you're seeing what I believe is the same problem/opportunity more from a type theoretic point of view (this is might be splitting hairs, sorry).

> 
> 1. `type ^ 0 = unit`
> 2. `type ^ 1 = type`
> 3. `type ^ 2 = type * type` (i.e., a binary tuple / pair)
> 4. `type ^ n = type * type * type * ...` (i.e., an n-ary tuple)
> 
> This perspective is quite neat, because applying it to the source language would mean that there needs to be no standalone array type (dedicated constructors for homogeneous tuples and tuple types will still be needed).
> 
> The problem: the current source language doesn't view type as just sets of values. Examples:
> 
> 1. `(UnionTy (IntTy) (UnionTy (IntTy) (FloatTy)))` != `(UnionTy (IntTy) (FloatTy))`
> 2. `(TupleTy (IntTy))` != `(IntTy)`
> 
> In both cases, both types represent the exact same set of values, but the types are considered not equal. Number 1 was a deliberate choice (to allow for better composition, especially once type inference, type variables, and type parameters are added), number 2 less so.

Yup, I remember that discussion and decision and I think it was quite principled at the time, and still think it holds up today.

As I said above, I think our thinking is probably converging towards a similar place, we're likely missing some type theory bits. I'm reminded of how, in linear algebra or vector mathematics, the scalar `1` and the matrix or vector `[1]` are almost equivalent, but it does break down for various operations. We might need to look at things from an operational perspective (operation plus data) rather than strictly a data type perspective, that seems to hint towards the benefits of a more structural first view of types (as opposed to nominal first).

With that said, perhaps there is an operation, like "normalize", or "reduce" but I hope you get the idea, that takes the LHS of 1 and products the RHS of 1, meaning a `normalized_equality`, or the like, could exist where 1 becomes true for that new equality, but not the current equality operator being used in 1 at present.

> Nothing of the above is a blocker for this PR, but it does hint at the current type theory not being formal / principled enough.

languages/source.nim
```diff
@@ -291,6 +291,28 @@ const lang* = language:
...
+        # it's a duplicate; skip
+        unionOfAux(typ_2, accum)
+      else:
+        case accum
+        of TupleTy(*typ_3):
+          unionOfAux(typ_2, TupleTy(...typ_3, typ_1))
+    else:
+      accum
+
+  func unionOf(list: *typ) -> typ =
```
saem, line 309: Ah, so this "flattens the union", or canonicalizes/normalizes it, meaning an operation resulting in the smallest type representation that still matches the same types as the previous one.

It _might_ be worth renaming to `normalizeUnion`, or derivative, but I'm not convinced and/or don't feel too strongly about it.

languages/source.nim
```diff
@@ -775,10 +802,20 @@ const lang* = language:
       let val_2 = val_1[n_1]
       conclusion At(array(*val_1), IntVal(n_1)), val_2
 
+    rule "E-at-tuple":
+      condition 0 <= n_1
```
saem, line 806: Perhaps it's just me, but I still find this "backwards", and need to mentally flip this to `n_1 >= 0` in order to understand it.

Soft comment, doesn't need to change, just wondering if you trip up on this too?

saem (approved):
LGTM

Minor comments, but nothing that's a blocker.

languages/source.nim
zerbina: > Ah, so this "flattens the union", or canonicalizes/normalizes it, meaning an operation resulting in the smallest type representation that still matches the same types as the previous one.

Hmm, not quite. I think a more accurate description would be "treat this unordered *list* of types as a *set* (meaning no duplicates) and create a `UnionTy` from it". So it's not `union -> union`, but rather `bag of types -> maybe union`.

I'd have preferred the "smallest type" operation, but that'd be problematic for multiple reasons (one of them outlined [here](https://github.com/nim-works/phy/pull/135#issuecomment-2746413647)).

languages/source.nim
zerbina: > Soft comment, doesn't need to change, just wondering if you trip up on this too?

Yea, I still trip on it sometimes, though I'm getting somewhat used to this notation, as "upper and lower bound for `x`" is often written `lb <= x <= ub` in the literature.

The only reason I'm using the "constant before variable" notation here is because the meta-language has no `>=` function :D. There's no reason for the meta-language to not have it -- the current notation just didn't bother me enough to PR `>` and `>=`.

zerbina:
> There are a few major points around "vector types" (array, tuple, etc), as opposed to "scalar" (bool, int, char, etc), that "vector types" have:

I like the idea of calling aggregate types like arrays, tuples, etc. "vector types". Some additional thoughts:
* I think it may be better to only call types where the elements are ordered by their *position* (so arrays, tuples, etc.), since - from what I've seen - dimensions of vectors are understood to be ordered
* "order" is also a property I intuitively thought of, but unordered vs. ordered might be too broad, as it gives rise to the questions "order between what? (position, value, name, etc.) and "what type of order? (partial, total, etc.)". Maybe this is not important/relevant in practice

> We might need to look at things from an operational perspective (operation plus data) rather than strictly a data type perspective

I agree.

> With that said, perhaps there is an operation, like "normalize", or "reduce" but I hope you get the idea, that takes the LHS of 1 and products the RHS of 1, meaning a `normalized_equality`, or the like, could exist where 1 becomes true for that new equality, but not the current equality operator being used in 1 at present.

I'm not sure I understand what you mean. Do you mean an equality operator that - in the context of unions - "normalizes" the unions prior to comparison?

saem:
> > There are a few major points around "vector types" (array, tuple, etc), as opposed to "scalar" (bool, int, char, etc), that "vector types" have:
> 
> I like the idea of calling aggregate types like arrays, tuples, etc. "vector types". Some additional thoughts:
> 
> * I think it may be better to only call types where the elements are ordered by their _position_ (so arrays, tuples, etc.), since - from what I've seen - dimensions of vectors are understood to be ordered
> * "order" is also a property I intuitively thought of, but unordered vs. ordered might be too broad, as it gives rise to the questions "order between what? (position, value, name, etc.) and "what type of order? (partial, total, etc.)". Maybe this is not important/relevant in practice

Having "vector types" as a name for types that are ordered (where indices/field order indicate data/memory layout order, forgiving initial and/or trailing offsets or padding/alignment) seems entirely reasonable. That is to say, I do agree that the intuitive definition of vector from math, 3d graphics, etc... all indicate an ordering of dimensions.

The alternative/extra names I came up with was "row types" for both ordered and unordered (key/fields), where "vector types" and "record types" would be subsets taking ordered and unordered variants under them.

> > With that said, perhaps there is an operation, like "normalize", or "reduce" but I hope you get the idea, that takes the LHS of 1 and [produces] the RHS of 1, meaning a `normalized_equality`, or the like, could exist where 1 becomes true for that new equality, but not the current equality operator being used in 1 at present.
> 
> I'm not sure I understand what you mean. Do you mean an equality operator that - in the context of unions - "normalizes" the unions prior to comparison?

Yup, a special equality operator which "normalizes" all operands (types) prior to comparison is what I meant.

PR #136, opened 2025-03-24
spec: make dynamic semantics of `With` correct

zerbina:
## Summary

* make `With` semantics match the intended one
* make the meta-language `and` and `or` operators work properly

## Details

* add the built-in `updated` function to the meta language, for
  replacing list elements
* translate the `and` and `or` operators into `if` expressions during
  analysis, in order to achieve the short-circuit semantics
* use `updated` to make the `With` reduction rules work as intended
* move the `With` rules to the `pReducesTo` inductive relation, as
  they're context free

---

## Notes For Reviewers
* should wait with merging until #119 is merged
* more progress towards a correct source language definition


languages/source.nim
```diff
@@ -782,6 +782,12 @@ const lang* = language:
       condition n_1 < 0 or len(val_1) <= n_1
       conclusion At(array(*val_1), IntVal(n_1)), Unreachable()
 
+    rule "E-with-array":
+      where val_3, array(for y in updated(val_1, n_1, val_2): y)
```
saem, line 786: Sorry, I still struggle with `With`, this is my current mental model, I'm not sure where the error is just yet, likely I don't understand `val_3`:

* `val_1` is the array or tuple
* `n_1` is the index
* `val_2` is the new value
* `val_3` is the updated array or tuple?

saem (approved):

languages/source.nim
zerbina: No problem. Your mental model is almost correct, except for `val_1`, which is the array's or tuple's elements and not the array/tuple itself.

---

I think what might be the source of (or at least a contributor to) your confusion is how `rule`s are written/read. In terms of flow, they're written like deduction/inference rules are in the literature.

I'll go through how to read and interpret the rule:
```nim
inductive pReducesTo(inp e, out e):
  ...
  rule "E-with-array":
    where val_3, ...
    conclusion With(array(*val_1), n_1, val_2),
               val_3
```

The `inp`(ut) and `out`(put) specifications inform the shape of the `conclusion`s and how they're interpreted.

The first thing that you want to look at is the `conclusion`, more specifically the expressions in *input* positions (which, in this case, is the `With(...)` expression). These are patterns that (may) bind meta-language values to identifiers (`val_1`, `n_1`, and `val_2`, in this case), which are then available to the rest of the rule's body.

If any of the patterns don't match the inputs to the relation, the rule doesn't apply.

Then continue with reading from the top of the rule down to the conclusion. `where x, y` is means "match `y` with pattern `x` and fail the rule if there's no match". In this case, the `where` could also be written as `let val_3 = array(...)`.

The last thing to look at is the expression(s) in the output position of the `conclusion`. These are not patterns, just normal meta-language expressions.

A possible textual representation of the rule would be the following:

$$
\frac{val_3 = (array\ ...)}{(\textbf{With}\ (\textbf{array}\ val_1\star )\ n_1\ val_2 ) \rightarrow val_3 }
$$

PR #138, opened 2025-03-27
builtin: fix a compilation error

zerbina:
Fix the `updated` implementation not compiling due to still assuming
that numeric values are represented as `Int128`.


zerbina:
@saem: The compilation failure sneaked into the main branch due to #136 being merged after #119, but without the main branch being merged into it first.

I've enabled the "require branch to be up-to-date" requirement to prevent such thing from happening in the future. When I find the time, I'll migrate the project to merge queues.

PR #139, opened 2025-03-29
lang: add record types

zerbina:
## Summary

Add record types, record constructors, and record type constructors to
the source language. Record types are similar to tuples, with the
difference that the fields are looked up by name, not by position.

### Language Definition

* add the syntax for records. The type constructor uses `RecordTy`, the
  value constructor `RecordCons`, and the elimination re-uses
  `FieldAccess`, but with an identifier
* add the static and dynamic semantics for records. They're largely the
  same as those for tuples, but using name lookup instead of positional
  lookup

### Meta Language

* make nested constructors (i.e., constructors appearing within
  constructors in data type definitions) proper ones that can be used
  standalone
* report a proper error when using undeclared identifiers in patterns

### Source2IL

* implement records as defined by the source language specification

### Tests

* add tests for all aspects of records (construction, access, etc.)
* render `Ident("x")` as just `x` in the output of `spectest`, so that
  the record tests' output specification can be shorter

Multiple new tests are marked as known issues in `spectest`:
* the `TypeDecl`-using ones need proper top-level code semantics to work
* `t22_record_type_equality_1.test` works, but the output comparison
  only compares the rendered results (which differ), not the actual
  values (which are equal)

### DIFF
--- a/languages/source.nim
+++ b/languages/source.nim
@@ -19,15 +19,21 @@ const lang* = language:
   #       numbers, not *integer* numbers
   typ ident: Ident(string)
   alias x, ident
+  typ eindex:
+    # element index
+    IntVal(z)
+    Ident(string)
+
   typ expr:
     Ident(string)
     IntVal(z)
     FloatVal(r)
     TupleCons(*expr)
+    RecordCons(+Field(ident, expr))
     Seq(texpr, *expr)
     Seq(StringVal(string))
     Call(+expr)
-    FieldAccess(expr, IntVal(z))
+    FieldAccess(expr, eindex)
     At(expr, expr)
     As(expr, texpr)
     And(expr, expr)
@@ -53,10 +59,12 @@ const lang* = language:
     `proc`(typ, *[x, typ], e)
 
     With(e, n, e) # return a tuple/array value with the n-th element replaced
+    With(e, string, e)
     Frame(typ, e) # a special expression for assisting with evaluation
 
   alias e, expr
 
+  typ fieldDecl: Field(ident, texpr)
   typ texpr:
     Ident(string)
     VoidTy()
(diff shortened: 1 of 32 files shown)


zerbina:
The source language's record types are similar to Nim/NimSkull's named tuples (i.e., `tuple[x: T, y: U, ...]`), with the difference that they:
* have no type relationship with tuples
* are unordered (i.e., order of fields doesn't matter for equality).

I thought about adding a subtype relationship such that `(RecordTy (Field x (IntTy)) (Field y (FloatTy))) <: (RecordTy (Field x (IntTy)))` (row polymorphism?), and while implementable with relative ease (in a naive manner), this could spell trouble for the semantics and implementation of pointers and var / mutable parameters (once they or something similar gets added to the language), which is why I decided against going down this route for now.

saem:
> The source language's record types are similar to Nim/NimSkull's named tuples (i.e., `tuple[x: T, y: U, ...]`), with the difference that they:
> 
> * have no type relationship with tuples
> * are unordered (i.e., order of fields doesn't matter for equality).
> 
> I thought about adding a subtype relationship such that `(RecordTy (Field x (IntTy)) (Field y (FloatTy))) <: (RecordTy (Field x (IntTy)))` (row polymorphism?), and while implementable with relative ease (in a naive manner), this could spell trouble for the semantics and implementation of pointers and var / mutable parameters (once they or something similar gets added to the language), which is why I decided against going down this route for now.

I think holding off on naive row polymorphism of this variety is the right call. From my reading of row polymorphism a while back there are many levels, the levels defined had a functional programming bent which tended to gloss over some more "imperative style" concerns.

We probably want to hold off on implementing row polymorphism until we have a mental map of the various kinds and can gauge which level we want to support.

saem (approved):
LGTM

Just one thought: apologies if I didn't catch it, but could we optimize the case of a `unit` field, where we can have such fields but we don't take up any storage for them (as the value is always known)?

zerbina:
> Just one thought: apologies if I didn't catch it, but could we optimize the case of a `unit` field, where we can have such fields but we don't take up any storage for them (as the value is always known)?

Yea, we could, and further down the line we should, but right now it'd complicate `source2il`, so I'm holding off on this optimization until it gets easier to implement.

PR #141, opened 2025-04-03
lang: add float multiplication and division

zerbina:
## Summary

Add the `*` and `/` operators for float values. Their operational
semantics are defined according to the IEEE 754 standard.

## Details

Both the definition in `source` and implementation in `source2il` are
straightforward. Tests are added to make sure multiplication and
division works, especially when involving non-finite values.

Integer and floating-point division are different enough to warrant
using separate identifiers (`/` vs. `div`). This also makes it possible
to add a `/` overload accepting integer values, in the future.

Division is defined in terms of a rational number division + to-float
conversion, as this is easier to understand than a division algorithm
using only integer numbers.

### Misc

Fix a small issue with the float addition overflow tests; they were
using constants that are outside the finite 64-bit floating point range,
meaning that the tests were effectively testing `inf` + finite. The
constants are changed to the largest possible finite values.

### DIFF
--- a/languages/source.nim
+++ b/languages/source.nim
@@ -334,8 +334,8 @@ const lang* = language:
       conclusion C_1, SeqTy(texpr_1), SeqTy(typ_1)
 
   def builtIn,
-    {"==", "<=", "<", "+", "-", "*", "div", "mod", "true", "false", "write",
-     "writeErr", "readFile"}
+    {"==", "<=", "<", "+", "-", "*", "/", "div", "mod", "true", "false",
+     "write", "writeErr", "readFile"}
 
   func stripMut(t: typ) -> typ =
     case t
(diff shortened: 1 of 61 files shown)


zerbina:
The original plan was to also add the remainder operator, but the IL has no support for it, and since the floating-point remainder is a not-so-commonly-used operation, I didn't include it.

The operator is definitely going to be added to the language at one point, just not now.

languages/source.nim
```diff
@@ -941,6 +969,56 @@ const lang* = language:
...
+    of Inf(bool), Inf(bool):
+      Nan
+    of Zero(bool), Zero(bool):
+      Nan
+    of Inf(bool_1), Finite(bool_2, n, z):
+      Inf(bool_1 xor bool_2)
+    of Finite(bool_1, n, z), Inf(bool_2):
+      Zero(bool_1 xor bool_2)
+    of Inf(bool_1), Zero(bool_2):
+      Inf(bool_1 xor bool_2)
```
saem, line 1011: TIL: that this is because zero in most cases is due to underflows, due to the nature of the spec, while inf is mostly overflows, so dividing by zero is seen as an overflow/underflow, that's also why we take the zero's sign into account.

crazy

saem (approved):

PR #142, opened 2025-04-04
lang: add array type

zerbina:
## Summary

Add an array type to the source language, which represents a
homogeneous fixed-size vector type. Array values are constructed with
`ArrayCons`.

## Details

Arrays have significant overlap with homogeneous tuples, to the point
that they are essentially isomorphic (but they aren't considered as
such by the language, at the moment), most likely requiring a rework in
the near future.

For the moment, `ArrayTy` provides a more convenient way (when compared
to `TupleTy`) to express homogeneous vectors with many elements.
Compared to `TupleCons`, `ArrayCons` requires the elements' types to be
homogeneous, or that they at least have a common ancestor type.

### DIFF
--- a/languages/source.nim
+++ b/languages/source.nim
@@ -50,6 +50,7 @@ const lang* = language:
     Ident(string)
     IntVal(z)
     FloatVal(float)
+    ArrayCons(+expr)
     TupleCons(*expr)
     RecordCons(+Field(ident, expr))
     Seq(texpr, *expr)
@@ -79,6 +80,7 @@ const lang* = language:
     char(z) # UTF-8 byte
     loc(z) # first-class location
     `array`(*val)
+    # TODO: ^^ use the `Seq` constructor instead; it also preserves the type
     `proc`(typ, *[x, typ], e)
 
     With(e, n, e) # return a tuple/array value with the n-th element replaced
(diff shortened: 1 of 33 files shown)


zerbina:
My initial thinking was that, with tuples effectively supporting all operations that integer-indexed arrays do (not yet, but they will), there's no point in having a dedicated type for arrays. Expressing large homogeneous tuples with `TupleTy` is effectively not possible, so my idea was to have an `Array` type constructor that's a shorthand for `TupleTy` with N elements of the same type.

I changed my mind mainly because of zero-sized arrays. My plan is for raw character arrays to be typed as fixed-sized homogeneous vector types, meaning that `""` needs a type too, which is where zero-sized arrays would come in.

The problem: when the array type constructor is `ArrayTy(i, T)`, zero-sized array have phantom type information, which the structurally and operationally equivalent `unit` does not, meaning that arrays and tuple are not entirely identical, at least when phantom type information can be retrieved (it currently can't be, but that'll likely be subject to change).

My current thinking is that arrays should be distinct types, but with non-zero-sized arrays and homogeneous tuples being *isomorphic*. That is, arrays and homogeneous tuples sharing the same element count and type should be used interchangeably. Due to a shortcoming of the language definition, I have not implemented this yet, and I'm also not entirely sure whether this really is a good idea.

saem (approved):

PR #144, opened 2025-04-08
interpreter: start with using tail calls

zerbina:
## Summary

Start with using tail calls for the continuations in the `interpret`
procedure. This prevents stack overflows when interpreting functions
with very complex bodies.

## Details

So far, continuations were just arbitrary closures, but now they're
`.tailcall` procedures. Anonymous `.tailcall` procedure cannot capture
outer variables, so captures have to be handled manually. A small set of
simple macro helpers (the `cps` module) is introduced to eliminate the
boilerplate code this would entail.

The main change is made out of two parts:
1. the lambda sugar is replaced with `proc(...): ...` that use the
   `.cont` calling convention (which is a macro pragma)
2. where sensible and necessary, procedure parameters use `sink`, and
   owning locals and parameter are destroyed manually, using the `drop`
   helper. The inner procedures in `interpret` are also turned into
   `.tailcall` procedures, so that the the tail-call chain is not
   interrupted

The `cps` macros are kept as simple as possible, performing no analysis
of the procedure bodies. Continuations have value semantics and
capture everything by value by default.

Since closure environments are now only created when really needed, with
automatic moves used wherever possible, this change also noticeably
improves run-time performance of the interpreter.

The required NimSkull compiler version is bumped to the most recent one,
as it include a bugfix necessary for the changes to compile.

---

## Notes For Reviewers

* this is the first step in using tail calls throughout the
  `interpreter` module; next up is updating the pattern matching
* since only the core interpreter loop uses tail-calls at the moment,
  stack overflows are still possible with long patterns or deep function
  recursion
* for review, I recommend first taking a look at the `cps` module

### DIFF
--- a/spec/cps.nim
+++ b/spec/cps.nim
@@ -27,6 +27,8 @@ proc `=copy`(dst: var CellRef, src: CellRef) =
       dst.p = src.p.copy(src.p)
 
 proc take[T](c: sink CellRef): T {.inline.} =
+  ## Returns the value from the given cell, `c`, which has to have dynamic
+  ## type `T`.
   move (Cell[T])(c.p).val
 
 proc copyImpl[T](x: CellBase): CellBase =
--- a/spec/cps.nim
+++ b/spec/cps.nim
@@ -1,0 +1,150 @@
+## Implements some macros and types to make writing code in continuation-
+## passing style easier.
+
+import std/macros
+
+{.experimental: "callOperator".}
+
+type
+  CellBase = ref object of RootObj
+    ## Header for a heap cell.
+    copy: (proc(x: CellBase): CellBase {.nimcall, raises: [].})
+  Cell[T] = ref object of CellBase
+    val: T
+
+  CellRef = object
+    ## Type-erased managed unique pointer to a heap cell. Needed for storing
+    ## the environment for continuations.
+    p: CellBase
+
+# --- cell implementation
+
+proc `=copy`(dst: var CellRef, src: CellRef) =
+  if dst.p != src.p: # do nothing for self assignments
+    if src.p.isNil:
+      dst.p = nil
+    else:
+      dst.p = src.p.copy(src.p)
+
+proc take[T](c: sink CellRef): T {.inline.} =
+  move (Cell[T])(c.p).val
+
+proc copyImpl[T](x: CellBase): CellBase =
+  Cell[T](val: Cell[T](x).val, copy: x.copy)
+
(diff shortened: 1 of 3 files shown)


zerbina:
The PR is ready for review -- the compilation failure is caused by a NimSkull compiler bug that requires fixing.

spec/cps.nim
```diff
@@ -0,0 +1,150 @@
...
+# --- cell implementation
+
+proc `=copy`(dst: var CellRef, src: CellRef) =
+  if dst.p != src.p: # do nothing for self assignments
+    if src.p.isNil:
+      dst.p = nil
+    else:
+      dst.p = src.p.copy(src.p)
+
+proc take[T](c: sink CellRef): T {.inline.} =
```
saem, line 29:
```suggestion
proc take[T](c: sink CellRef): T {.inline.} =
  ## returns the value held by the cell, `c`, as type `T`
```

I was explaining the code to myself and I figured I'd write it down here

spec/cps.nim
```diff
@@ -0,0 +1,150 @@
...
+  ## Storage wise, a continuation is a tuple.
+  typ.expectKind nnkProcTy
+
+  let ntyp = copyNimTree(typ)
+  ntyp.pragma.add ident"tailcall"
+  ntyp[0].add newIdentDefs(ident"env", paramType())
+
+  result = nnkTupleConstr.newTree(ntyp, bindSym"CellRef")
+
+  # for reasons that are beyond silly (the compiler analyzes the proc type
```
saem, line 131: D'oh!

spec/interpreter.nim
```diff
@@ -287,9 +296,15 @@ template catch(c: var Context, body, els: untyped) =
...
     els
 
-proc interpret*(c; lang; n: Node, then: Next): Node
+proc interpret*(c; lang; n: Node, then: sink Next): Node {.tailcall.}
+
+proc eval(c; lang; n: Node): Node =
+  ## Evaluates `n`, returning the result or raising an exception. Important:
+  ## this breaks the continuation chain.
+  interpret(c, lang, n,
+    proc(c; lang; val: sink Node): Node {.cont.} = val)
```
saem, line 305: This is a nice bit of cleanup

saem (approved):
LGTM

(minor suggestion if you want to include it)

spec/cps.nim
zerbina: I've used a slightly adjusted version of your suggestions, putting emphasis on the value being required to have the correct type `T`.

PR #145, opened 2025-04-10
spec: properly define the module semantics

zerbina:
## Summary

Make the static and dynamic semantics for modules reflect the
originally intended behaviour.

## Details

* add the desugaring logic for declarations and modules
* add the internal-only `Letrec` (let recursive) abstract syntax,
  together with its static and dynamic semantics
* assign a type to modules (either `void`, or the return type of the
  trailing procedure) and require the trailing procedure, if any, to
  have zero parameters 
* add the dynamic semantics for modules; the procedure declarations are
  reduced to a nested series of `Letrec` expressions

Desugaring a `Module` into a series of `Letrec` expressions keeps
the dynamic semantics simpler, also getting around `Module` either
becoming an expression, or `cstep` and friends having to handle the
`module` type.

Some fixes and additions to the meta language (type checker and
interpreter) are required:
* unpacking in constructions can now use group unpacking (i.e.,
  `...[a, b, c]`)
* match patterns (in `of` branches and `where`) now have to be a
  subtype of or equal to the matched expression's type, not the other
  way around

Finally, `spectest` is updated to handle top-level code (by using the
appropriate relations/functions), with the test execution logic being
slightly refactored in the process.


zerbina:
In the meta language, types are effectively sets of values. When an identifier has a type, it means that the identifier stands for a value part of that type (set).

This wasn't clear to me from the beginning, as I started with defining operations and only afterwards tried to formalize the semantics somewhat, which is also the reason for the nomenclature being inconsistent throughout the type checker.

A few issues have accumulated by now, and I'll likely have to do a more thorough refactor of `langdefs` soon.

---

Regarding how the top-level semantics are defined (i.e., via the new `Letrec` construct), I think this PR marks the final departure of having the static and dynamic semantics be defined directly for the source language.

Initially, my goal was for the definition to use no helper abstract syntax, so that one doesn't have to effectively juggle two languages when developing the semantics. If the language were a lot simpler, this would have worked out, but it's not, and not cleanly separating the two languages makes the definition confusing, in my opinion.

Therefore, the revised direction is that there's a *core language* separate from but related to the source language. The core language should still share as much of abstract syntax from the source language as possible, however, and going from the source to the core language must never involve more than a pure syntax transformation.

Once the reference implementation works, and some of the needed missing meta language features are implemented, I'll split the syntax and semantics of the core into its own module (`core.nim`) -- `source.nim` will only contain the source language's abstract syntax + the source-to-core translation (i.e., the not-so-aptly named `desugar`).

languages/source.nim
```diff
@@ -513,6 +530,21 @@ const lang* = language:
       premise mtypes(C_2, e_2, typ_2)
       conclusion C_1, Let(Ident(string_1), e_1, e_2), typ_2
 
+    rule "S-letrec":
```
saem, line 533: If I'm understanding this correctly (big if), this looks to be similar to ocaml's `let rec'?

In this case, the initializer expression (`e_2`) could recurse (call the procedure named `Ident(string_1)`)

saem (approved):

languages/source.nim
zerbina: > this looks to be similar to ocaml's `let rec'?

I'm not too familiar with OCaml, but reading through the documentation, the constructs seem largely identical.

To give another point of reference, the `Letrec` construct here does the same thing as `(letrec [(x y)] ...)`, but with the requirement that `y` *must* be a lambda expression.

`Letrec` is meant to be part of the *core language*, and there are currently no plans to make it available in the source language (this can still change, of course). The idea is to desugar all potentially-recursive procedure definitions into `Letrec`, which is easier to reduce/type, as the scope of the identifier is explicit (i.e., it only exists within the initializer and body).

For example (using NimSkull syntax for convenience):
```nim
proc a(): int =
  proc b(x, y: int): int =
    b(y, x)
  b(1, 2)
discard a()
```
gets desugared into:
```nim
letrec a = proc(): int =
  letrec b = proc(x, y: int): int =
    b(y, x)
  b(1, 2)
discard a()
```

saem: Thanks for the explanation, @zerbina -- that very much looks to be the same as the ocaml one, which is pretty neat that you ended up in the same place. :)

To prove to myself that I understand it, but also to anyone reading in the future, the key point in @zerbina's example above is that the lambda assigned to `b` in `letrec b` has in its scope/environment the binding `b`, such that the call `b(y, x)` can be called without resulting in an undefined error. If it was simple a `let`, then `b` wouldn't be available within that scope/environment.

PR #146, opened 2025-04-13
WIP: add an `L3` -> LLVM IR translation pass

zerbina:
Add a pass that translates `L3` code into LLVM IR, making it possible to compile source language programs and NimSkull program's compiled via `skully` to native executables via LLVM.

At the moment, the pass is biased towards code produced by `skully`.

Instead of using the LLVM API to create the LLVM IR and compile it into an executable, the pass simply emits textual LLVM IR, which has a number of benefits:
* no bindings to the LLVM API are needed
* `phy` doesn't depend on a dev version of LLVM; LLVM doesn't have to be built from source to compile `phy`
* the textual IR is much more stable than the LLVM API

In the future, the pass could move to emit LLVM bitcode, which would speed up the pass and produce smaller artifacts.

---

## To-Do
* [ ] fix the upstream (i.e., NimSkull) bugs that make `skully` produce incorrect code
* [ ] move the `skully` fixes into a separate PR
* [ ] move the `Ptr` type introduction to a separate PR
* [ ] split `Conv` into multiple operators (separate PR)
* [ ] properly implement exception handling
* [ ] implement `Select` support
* [ ] split up `runtime.c`  
* [ ] test `skully` + LLVM pass in the CI
* [ ] write a proper commit message

### DIFF
--- a/languages/lang3.md
+++ b/languages/lang3.md
@@ -17,7 +17,9 @@ typedesc -= (Blob <int> <int>)
 typedesc += (Record size:<int> align:<int> (Field <int> <type>)+)
           | (Union  size:<int> align:<int> <type>+)
           | (Array  size:<int> align:<int> count:<int> <type>)
+          | (Ptr) # XXX: temporary solution
 
+type += (Ptr) # XXX: temporary solution
 type -= (Blob <int> <int>)
 ```
 
--- a/passes/pass_aggregateParams.nim
+++ b/passes/pass_aggregateParams.nim
@@ -100,7 +100,7 @@ proc typeof(c; tree; n): Node =
     unreachable()
 
 proc newLocal(c; typ: Node): Node =
-  assert typ.kind in {Type, UInt, Int, Float}
+  assert typ.kind in {Type, UInt, Int, Float, Ptr}
   result = Node(kind: Local, val: uint32(c.firstTemp + c.temps.len))
   c.temps.add typ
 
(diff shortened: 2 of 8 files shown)


zerbina:
Locally, I'm able to compile `phy.nim` into a working native executable, like so:
```sh
> skully phy/phy.nim phy.txt
> phy --source:L25 --target:llvm phy.txt
> llvm-as -o phy.bc out.ll
> llc -filetype=native -o phy.o phy.bc
> clang phy.obj runtime.c 
```

It's also possible to do combine the LLVM steps by using (stock) clang, but this routinely resulted in clang crashing, at least on my system.

The above is not reproducible with a stock NimSkull yet, as there are multiple bugs in the NimSkull compiler where incorrectly typed MIR code is emitted. `skully` doesn't perform any type correction (nor should it), thus leading to incorrectly typed L3 code, and since the LLVM IR is quite strongly typed (when compared to, e.g., the Phy VM bytecode), this presently results in `llvm-as` rejecting the input.

---

This PR will most likely stay open for quite some time, as there's no rush to finish it. It's also not clear whether it should be merged at all, or whether it's better for the LLVM pass to be implemented in a separate repository.

I mainly wanted to experiment with how easy/possible it is to implement an to-LLVM translation pass with the current compiler.

zerbina:
I've merged the PR with upstream and also added a test for the code generator to the CI.

There's still one issue remaining in the NimSkull compiler, namely, that the `pred` and `succ` translation in `mirgen` produces incorrect MIR code, which shows up here as a type error from the LLVM assembler.

Separately, I've concluded that the IL `Select` was a misfeature. I'll replace it with something akin to a branch-with-table, which is much simpler to implement, not only for the LLVM code generator, but also the VM code generator.

PR #147, opened 2025-04-16
spec: fix the call and return dynamic semantics

zerbina:
## Summary

Make the dynamic semantics regarding calls and `Return` match the
intended ones. This means that the code in most call-using tests is now
officially legal.

## Details

With the previous context definition `B = E | Frame(typ, B)`, `B` does
not match expressions such as `Frame(typ, Exprs(Frame(typ, ...)))`,
since `Exprs(Frame(typ, ...))` matches neither `E` nor `Frame(typ, B)`.

The `B` context is used in frame reduction (i.e., unwinding), meaning
that nested calls were mostly irreducible.

In addition, since `E` doesn't include `Frame`, reducible expressions
nested in a `Frame` were not reduced, meaning that reduction got stuck
for any call of a procedure with a non-trivial body.

To properly implement the `B` context, the meta language is extended to
support limited usage of plug-patterns (e.g., `E[...]`) in a context
(which is effectively a named or-pattern). This feature allows
defining `B` in a way such that it matches everything that `E` does,
plus `Frame`, without manually having to copy and alter the body of `E`.

Finally:
* the `Return` reduction rules are changed to only replace the
  enclosing local context with the returned value, without "unwinding"
  the frame
* a new rule is added that reduces a `Frame` with only a value as the
  body to the value (effectively "returning" it)


zerbina:
424 out of 442 tests are now officially legal, according to the language definition!

For those wondering what the purpose of `Frame` is: it's there to mark the start of a procedure invocation's context. Without it, `Return` wouldn't know what to reduce.

Regarding the meta-language `context` construct, the direction is for it to go back to being just a named pattern, admitting all patterns (including mutually recursive named patterns) also allowed elsewhere, except for binding ones. For this to work, the pattern matching implementation interpreter first has to be improved, which depends on delimited continuation support provided by the NimSkull -- the `cps`-supplied macros worked okay enough for the interpreter's main loop, but the pattern matching logic would become unreadable/unmaintainable (I tried).

saem (approved):

PR #152, opened 2025-04-24
interpreter: use tail-calls for list matching

zerbina:
## Summary

Make the routines involved in list pattern matching `.tailcall`
routines. Usage of complex list patterns can no longer cause a stack
overflow (by itself).

## Details

So that tail-calling is possible, all list matching routines now fully
use continuation-passing style, instead of a mixture between it and
direct style.

As a consequence, repetition matching becomes more complex, as it can
no longer inspect the return value of the "rest". Instead, it needs
access to both the "rest" *and* the return continuation of the "rest",
so that it can logically intercept the return.

The way the implementation now works means that a repetition pattern
nested in a group no longer works, but this was never explicitly
supported anyway and it's also disallowed by the meta-language's
semantic checker.

---

## Notes For Reviewers
* another step towards using tail calls everywhere in the interpreter

### DIFF
--- a/spec/interpreter.nim
+++ b/spec/interpreter.nim
@@ -131,85 +131,116 @@ proc wrap(a: sink Match): Match =
   result = a
   result.bindings = wrap(result.bindings)
 
-proc merge(a: var Match, b: Match) =
+proc merge(a: var Match, b: sink Match) =
   a.bindings.merge(b.bindings)
   a.ctx.add b.ctx
 
 proc matches(lang; pat: Node, term: Node): Match
 
-proc matchList(lang; pat: Node, term: Node): Match =
+proc matchList(lang; pat, term: Node): Match {.tailcall.} =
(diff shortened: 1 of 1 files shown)


zerbina:
In my first attempt at using tail calls for the list matching, I tried to model the repetition-related matching with try/except like control-flow, which proved to be too cumbersome to do with the current manual continuation creation.

While easier to implement and also a little more efficient, the second attempt (read, this PR) utilizes control-flow that is a bit of a mind bender. Once `suspend` (or something conceptually similar) is implemented in NimSkull, the code can be written in something that roughly resembles direct style again, which should help with reading and maintaining the code.

At present, the tail-callified list matching is slightly *slower* than the previous version. I haven't investigated why, but it also doesn't matter much at the moment -- what matters is that stack overflows are not possible.

saem (approved):

PR #153, opened 2025-04-25
interpreter: use tail-calls for pattern matching

zerbina:
## Summary

Make use of `.tailcall` for the remaining pattern matching logic.
Complex patterns can no longer cause stack overflow in the interpreter
by themselves.

## Details

The two `matches` routines are changed to use continuation-passing
style + tail calls, without anything else being changed.

### DIFF
--- a/spec/interpreter.nim
+++ b/spec/interpreter.nim
@@ -62,6 +62,8 @@ type
 
   Next = proc(c: var Context, lang: LangDef, val: sink Node): Node {.contcc.}
     ## type of a continuation
+  PNext = proc(lang: LangDef, m: sink Match): Match {.contcc.}
+    ## type of return continuation for pattern matching
 
 using
   c: var Context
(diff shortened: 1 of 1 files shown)


zerbina:
At least for me locally and when building with `-d:release`, running all specification tests now takes *twice* as long as it did before changing the pattern matching to use tail calls. I'm using a custom-built NimSkull compiler with some code-generator improvements, however, and the real numbers are likely to be worse.

With the NimSkull patch that fixes the tuple-unpacking memory leak (which I found by looking at the C code generated for tailcall procedures), the interpreter should become a bit faster again, as the bug also leads to unnecessary copies being made.

In any case, performance degraded enough to warrant profiling the interpreter and looking at where the extra time is spent. Skimming through the C code, I didn't spot any extraneous copies, so my current guess is that the tail-call elimination significantly inhibits optimization on the C side.

saem (approved):
Change looks good.

Even with it running slower, I'm going to merge this. I've merged the latest tuple leak fix, so after a version bump of nimskull we should get some performance back and hopefully things pick up with various changes.

PR #154, opened 2025-04-26
partially restore interpreter performance

zerbina:
## Summary

Use the most recent NimSkull version and selectively disable runtime
type checks. Interpreter performance is now closer to what it was prior
to the usage of tail calls for pattern matching, and memory usage no
longer grows indefinitely.

## Details

* using the most recent NimSkull version fixes both a severe memory leak
  and a lot unnecessary copies
* the object conversion type checks in `cps.take` made up ~25% of the
  execution time of `spectest`. Since the checks are unnecessary (the
  way `CellRef` is used makes dynamic type errors impossible), they're
  disabled

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -59,7 +59,7 @@ jobs:
       # use a pinned version in order to make CI runs reproducible
       - uses: nim-works/setup-nimskull@0.1.2
         with:
-          nimskull-version: "0.1.0-dev.21464"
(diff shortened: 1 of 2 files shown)


saem (approved):

zerbina:
In my local testing with a stock NimSkull compiler and `-d:release`, the interpreter is now only ~33% slower compared to when pattern matching was not fully CPS-ified, which is back in the acceptable range.

PR #156, opened 2025-04-28
introduce a dedicated AST type

zerbina:
The goal is to separate the generic node storage and traversal operations from the specific concept of an abstract syntax tree (=AST).

An AST requires some extra bits of information (e.g., metadata), but adding this to `PackedTree` would make the type it no longer applicable for use cases that want just a sequence of nodes and the associated operations without anything else.

While there currently are no such use cases, I can foresee that changing in the future, so splitting the two things makes sense.

As is, the current code is a dead end. Conceptually, the node tree operations (e.g.: `child`, `next`, etc.) do not care about the concrete storage type -- they only require an input type that's a random-access supporting container type whose element is a type with a mutable `kind` and `val` property, nothing more. This makes them an ideal candidate for static interfaces (i.e., `concept`), but said NimSkull feature is not yet stable enough.

I've tried to approximate the ideal solution (static interfaces) by making `PackedTree` type only represent the node storage and have it be parameterized with the node type, with `Ast` then embedding a `PackedTree` instance and providing a wrapper template for every `PackedTree` routine such that `Ast` can be used in the same way a `PackedTree` can be, but this is annoying to maintain and it also significantly hurts code clarity, since various routines for `Builder`, `ChangeSet`, etc. have to use unnecessarily broad interfaces.

Using inheritance would work a little better in practice, but is not really correct (an `Ast` cannot be used everywhere a `PackedTree` can be, for example, serialization), and it would likely also surface some NimSkull bugs/issues with argument subtyping and `iterator`s.

Progress on this PR needs to wait until NimSkull supports working static interfaces. In the meantime, `PackedTree` will have to become the `Ast` type.

### DIFF
--- a/passes/asts.nim
+++ b/passes/asts.nim
@@ -1,0 +1,141 @@
+## Implements a generic packed storage type for arbitrary abstract syntax
+## trees -- no restrictions regarding the grammar are made.
+
+import
+  passes/[literals, trees],
+  experimental/sexp
+
+type
+  AstNode*[T] = object
+    ## A single node. How the `val` field is interpreted depends on the kind.
+    ## For leaf nodes, the field's meaning is decided externally, while for
+    ## non-leaf nodes, `val` stores the number of child nodes the node has.
+    kind*: T
+    val*: uint32
+
+  Ast*[T] = object
+    ## A self contained abstract-syntax-tree (=AST).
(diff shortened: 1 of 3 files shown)


zerbina:
This refactor will be made obsolete with the eventual arrival of #179.

PR #157, opened 2025-05-01
phy: instrument pass execution

zerbina:
## Summary

Add the `--measure` switch to `phy`, which allows dumping internal
timings gathered for various steps. This is intended as a development
and optimization aid.

## Details

* add the `tracing` module, which implements a simple time-stamped-
  event log collection facility (`EventLog`)
* use `EventLog` to collect the start and end times for each internal
  pass, as well as `source2il` and VM code generation
* the timings are gathered from the log, rendered as a CSV list, and
  dumped to the standard output

To continue supporting compiling `phy` with `skully`, instrumentation is
disabled (at compile time), as the `times` and `monotimes` standard
library modules cannot be imported when using `--os:any` (which `skully`
always does).

---

## Notes For Reviewers
* `tracing` is effectively a heavily trimmed-down, more generic version of the NimSkull's `tracer` module

### DIFF
--- a/phy/phy.nim
+++ b/phy/phy.nim
@@ -61,6 +61,12 @@ import
 import passes/syntax except NodeKind
 import passes/syntax_source except NodeKind
 
+when hostOS != "any":
+  # unfortunately, skully doesn't support the times and monotimes module yet,
+  # meaning that tracing is not supported in a skully-built phy executable
+  import phy/tracing
+  import std/[tables, times]
+
 type
   Language = enum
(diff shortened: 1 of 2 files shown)


zerbina:
At the moment, data is dumped to the standard output, but in the future, it should be dumped to a file. If integrated with testing, a CI job could then gather the timings for all commits that land on `main` and then generate and publish a pretty graph based on the data, so that how the timings change can be tracked over time.

PR #159, opened 2025-05-03
make the IL `Call` grammar more regular

zerbina:
## Summary

Always have a type operand before the callee for every call syntax, no
longer making a grammatical distinction between static and dynamic
calls. `ProcVal` is removed.

## Details

`Call`, `CheckedCall`, and `CheckedCallAsgn` now always have a callee
type operand now, with the motivation being that:
1. it makes the procedure signature queryable directly at the callsite
   for static calls (not just for dynamic calls)
2. the grammar becomes simpler
3. some processing becomes more uniform 

Since the proc reference (`Proc`) vs. procedure value (`ProcVal`)
distinction is not useful anymore, the latter syntax is removed, with
`Proc` taking over its meaning.

The passes, `skully`, and `source2il` are adjusted to the aforementioned
changes. Tests using static calls (and their expected output) are
also updated to adhere to the new grammar. 

For the VM code generator, tests for indirect calls now have to use
local variables as the callee, as using the `Proc` directly now always
yields a static call.

---

## To-Do
* [ ] update the hand-written `L30` code in `source2il`

## Notes For Reviewers
* IL code size increase a bit, but I consider this to be outweighed by increase regularity
* passes that navigate/recompute types a lot benefit the most, as they no longer have to look up the `ProcDef` first


zerbina:
The hand-written `L30` in `source2il` needs to be updated, which is fairly annoying to do at the moment, as it would currently require manually passing all needed signature types to the `%` calls.

I'm waiting with the update, until the `L30` code is moved to dedicated modules read from disk, the latter which will make modifying the `L30` code a lot more comfortable.

zerbina:
I'm leaving this PR open for now, but it's effectively dead.

As part of the transition to using a proper nanopass framework (#179), the IL(s) will become a lot more regular in general (as that makes the passes easier to implement), and there's not really a point in refactoring code that is obsolete and (most likely) won't see much further usage.

PR #160, opened 2025-05-03
lang0: split `Conv` into multiple operators

zerbina:
## Summary

Add the `Zext`, `Sext`, `Trunc`, `Promote`, and `Demote` operators.
`Conv` still exists, but is now only used for int-to-float and float-
to-int conversions.

## Details

The previous `Conv` operator was modeled after the conversion operator
of NimSkull's MIR, making it fairly high-level and leaving too much
behaviour-related decision making to `pass0`.

It's split into multiple, lower-level operators, making the final
translation easier and also providing earlier transformation stages with
more precise conversion operators.

`skully` is changed to:
* use the new operators instead of `Conv` (where applicable)
* translate pointer-to-pointer conversions to IL no ops (as all
  pointers are translated to the same IL type)

Finally, tests for the new `L0` operators are added, except for `Conv`,
whose semantics aren't very well-defined at the moment.

---

## Notes For Reviewers

A preparation for #146, but the new operators should be helpful in general, especially once the source language starts to support numeric conversions.


saem (approved):
I personally prefer source  then destination (reading things as "from to") as the order, but if there is a good reason not to do that, I get it.

zerbina:
> I personally prefer source then destination (reading things as "from to") as the order, but if there is a good reason not to do that, I get it.

I chose the current order for reasons of regularity.

For most other operators that return a value (e.g., `Add`, `Sub`, `BitAnd`, etc.), the type of the result comes as the first operand (they don't have a second type operand), so when designing the initial `Conv` operator, I opted for having the destination type come first, so that `child(n, 0)` also returns the type operand representing the type of the operation.

There are some exceptions to the "first type operand is result type" rule, however, namely the comparison operators, `Not`, and `Call`.

PR #162, opened 2025-05-04
README with bare bones build instructions

saem:
Added a basic README for someone to get started. It's mostly to have
something that can be added to overtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,4 +16,5 @@ nim c --outdir:bin -r koch.nim all
 The above will:
 - compile `koch`, the build command
 - place that within the local `bin/` directory
-- finally, it'll build all programs in this repository
+- finally, it'll build all programs in this repository and place their
+  executables in the local `bin/` directory
--- a/README.md
+++ b/README.md
@@ -1,0 +1,20 @@
+# Building
+
+## Build Pre-requisites
+
+The build instructions assume you have NimSkull, latest preferred unless
+otherwise noted, installed and on the path.
+
+## Build Instructions
+
+This command will get everything built
+(N.B.: Windows users replace `nim` with `nim.exe` in the command below)
+
+```cmd
+nim c --outdir:bin -r koch.nim all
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Build Pre-requisites
+
+The build instructions assume you have NimSkull, latest preferred unless
+otherwise noted, installed and on the path.
+
+## Build Instructions
+
+This command will get everything built
+(N.B.: Windows users replace `nim` with `nim.exe` in the command below)
```
zerbina, line 11:
```suggestion
To get everything built, run the following from the command line
```
It's a little bit more clear and accessible this way, I think. Also, the `.exe` file extension for the executable is optional on Windoes - at least with the default cmd and PowerShell.

README.md
```diff
@@ -0,0 +1,20 @@
...
+
+```cmd
+nim c --outdir:bin -r koch.nim all
+```
+
+The above will:
+- compile `koch`, the build command
+- place that within the local `bin/` directory
+- finally, it'll build all programs in this repository
```
zerbina, line 20:
```suggestion
- finally, it'll build all programs in this repository and place their
  executables in the local `bin/` directory
```
It think it makes sense to also specify where the executable are placed (by default).

Also, this isn't entirely correct at the moment, since `skully` is currently skipped (for build performance reasons), but `kock all` really should do what the name implies, so I'll make a separate PR to make it work as advertised.

zerbina (approved):
Thank you. I've suggested two small improvements, but otherwise this looks good.

README.md
saem: Oh neat, I learned something new

README.md
saem: Yeah, I knew it didn't build `skully`, I left it this was because I figured we'd change that eventually (it gets faster or we add it).

PR #163, opened 2025-05-04
ci: read NimSkull version from file

zerbina:
## Summary

The target NimSkull version is now supplied via a file, so that it's
easier to spot and there's only a single place it needs to be updated.

## Details

So that the version only needs to be read once, a new CI job
(`context`) is added, which provides the required NimSkull version as
an output to other jobs.

---

## Notes For Reviewers

This should also help others with building `phy`, by making the target NimSkull version easy to find (no need to look through the workflow yaml).


saem (approved):

saem:
Actually, if you don't mind making a very tiny addition to this PR, could you edit the readme, where it says requirements, and link to the file so people can check if their nimskull is new enough?

Unfortunately, we can't embed it from the file without using a permalink (would go out of date, blech), ref: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet

zerbina:
I've used a link to the file at `main`'s head, as I *believe* I've seen those be rendered inline, too. Let's see whether they really do.

saem (approved):
perfect!

PR #165, opened 2025-05-07
vmmodules: support incremental loading

zerbina:
## Summary

Support VM modules being loaded incrementally (rather than all at once)
and fix a relocation bug.  

## Details

Previously, creating a VM program required all modules to be available,
with no way to load/link additional modules afterwards. Some use cases,
such as just-in-time compilation, require this however.

Changes related to module loading:
* the `link` procedure is replaced with `load`, which loads a single
  module into the VM environment, resolving imports where possible
* the `LinkTable` type is added for keeping track of the exports and
  unresolved symbols across module loads
* host procedure have to now be loaded via a separate procedure (also
  named `load`)
* the `get` procedure is added for querying a `LinkTable` instance
* a bug in `vmmodules.append` is fixed, where static calls weren't
  patched properly

All linking related state and logic is still fully separate from the
core VM.

The VM setup logic in the VM test runner, `phy`,  and the `repl` is
updated to use the loading procedure.

Finally, the VM test runner is extended to support creating multiple
VM modules, and this is used for expanding the import resolution tests.

---

## Notes For Reviewers
* a prerequisite for a better REPL (one that uses incremental compilation)
* a prerequisite for source-language-level modules
* a prerequisite for VM module memory regions (which are needed for global location support)


phy/repl.nim
```diff
@@ -107,7 +107,11 @@ proc process(ctx: var ModuleCtx, reporter: Reporter,
       unreachable("memory config invalid; there's probably a bug in source2il")
 
     var env = initVm(mem.total, mem.total)
-    link(env, hostProcedures(includeTest = false), [module])
+    var ltab = LinkTable()
```
saem, line 110: OK, I kinda see how we could have module definition swapping, using the `LinkTable` to link, and more importantly unlink/relink, procedures if we get updated definitions.

saem (approved):

phy/repl.nim
zerbina: Yeah, that would work. Not right now, as `load` doesn't touch existing exported procedures, but it's fairly easy to support.

PR #169, opened 2025-05-08
koch: replace `single` and `all` with `build`

zerbina:
## Summary

Replace the `single` and `all` commands with the more intuitive `build`
command. `build all` builds all programs (including `skully`), whereas
`build all-ws` builds all programs *except* `skully`.

## Details

The build logic in `skully` is restructured to be more generic. To make
lookup easier, the list of programs is now defined using a table, not
a sequence, with the program name used as the key.

For performance reasons, CI continues to build and test `skully`
separately from the other programs.

### DIFF
--- a/koch.nim
+++ b/koch.nim
@@ -12,8 +12,8 @@ Options:
   --nim:path                  use the specified NimSkull compiler
 
 Commands:
-  build <target> [<arg> ...]  builds a program with the given arguments. 'all'
-                              builds all programs
+  build <program> [<arg> ...] builds a program, using the given compiler args
+  build all [<arg> ...]       builds all programs
   generate <dir>              generates the various language-related modules
   build-defs                  verifies the language definitions and generates
                               the textual representation for them
--- a/koch.nim
+++ b/koch.nim
@@ -12,23 +12,22 @@ Options:
   --nim:path                  use the specified NimSkull compiler
 
 Commands:
-  all [args]                  builds all programs
-  single <name> [args]        builds the single program with the given name
-  generate [dir]              generates the various language-related modules
+  build <target> [<arg> ...]  builds a program with the given arguments. 'all'
+                              builds all programs
+  generate <dir>              generates the various language-related modules
   build-defs                  verifies the language definitions and generates
                               the textual representation for them
 """
-  Programs: seq[(string, string, bool, bool)] = @[
-    ("tester", "tools/tester.nim", true, true),
-    ("passtool", "tools/passtool/passtool.nim", true, true),
-    ("queryshell", "phy/queryshell.nim", true, true),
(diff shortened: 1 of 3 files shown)


koch.nim
```diff
@@ -12,23 +12,22 @@ Options:
...
-    ## program name, module path, whether the program doesn't depend on
-    ## generated modules, and whether the program is built with 'all'
+  Programs = {
+    "tester" : ("tools/tester.nim", true),
+    "passtool" : ("tools/passtool/passtool.nim", true),
+    "queryshell" : ("phy/queryshell.nim", true),
+    "repl" : ("phy/repl.nim", false),
+    "phy" : ("phy/phy.nim", false),
+    "skully" : ("skully/skully.nim", true),
+  }.toOrderedTable
```
saem, line 28:
```suggestion
  type
    Program = enum
      testerProg = "tester"
      passtoolProg = "passtool"
      queryshellProg = "queryshell"
      replProg = "repl"
      phyProg = "phy"
      skullProg = "skully"

  Programs: = [
    testerProg : ("tools/tester.nim", true),
    passtoolProg : ("tools/passtool/passtool.nim", true),
    queryshellProg : ("phy/queryshell.nim", true),
    replProg : ("phy/repl.nim", false),
    phyProg : ("phy/phy.nim", false),
    skullyProg : ("skully/skully.nim", true),
  ]
```

You might have already considered this, in which case feel free to disregard, I just lean towards more static things.

saem (approved):
A couple of soft suggestions.

If you like the array one but don't want to make the change I get, I can poke at it later if it makes sense.

koch.nim
```diff
@@ -12,23 +12,22 @@ Options:
   --nim:path                  use the specified NimSkull compiler
 
 Commands:
-  all [args]                  builds all programs
-  single <name> [args]        builds the single program with the given name
-  generate [dir]              generates the various language-related modules
+  build <target> [<arg> ...]  builds a program with the given arguments. 'all'
```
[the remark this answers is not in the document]
zerbina: Yea, build systems tend to refer to the concept as "target" (and I do think it applies here, as not only just a program may be built, but also its dependencies), but not everybody is necessary familiar with the term, so I think you're right in that using `program` makes it easier to understand in general.

I'll split it into two help entries.

koch.nim
zerbina: I don't think it makes sense here, as `koch` doesn't need to refer to individual programs itself (except for `skully`, but that's an exception) and all access of the table uses a user-supplied string value anyway.

If `koch` grows more special cases for individual programs (though I think it shouldn't), then it could be a good idea to use an enum, but right now, I don't see the benefit of using one.

PR #170, opened 2025-05-10
add an IL with global locations

zerbina:
## Summary

* add the `L6` IL, providing mutable global locations
* extend the `L0` IL with memory region support (`Data`)

## Details

* add the `Data` initializer for globals to the `L0` IL, which provides
  a convenient interface for VM module memory regions, initializer, and
  relocations
* implement the `Data` initializer in `pass0` and add tests for it
* add the `L6` IL, which extends the `L5` IL and is the new basis for
  the `L25` IL. It introduces `GlobalLoc` definitions, which define
  mutable global locations
* implement an `L6` -> `L5` lowering pass (`pass_globalsToPointer`) and
  add tests for it

Past the `L6`, a distinction between global locations (`GlobalLoc`) and
global constants (`GlobalDef`) still has to be kept for now, as the
legacy memory configuration still relies on global constants.

### DIFF
--- a/languages/lang6.md
+++ b/languages/lang6.md
@@ -1,0 +1,35 @@
+## L6 Language
+
+```grammar
+.extends lang5
+```
+
+It's now possible to define global *variables*, which are locations that can be
+assigned to, read from, and have their address taken.
+
+Non-variables continue to works as they did previously.
+
+```grammar
+globalInit -= <dataInit>
+globaldef += (GlobalLoc <type> <dataInit>)
+
+dataInit -= (Data <int> <int>)
+          | (Data <int> (StringVal <string>))
+dataInit += (Data <type>)
+          | (Data <type> (StringVal <string>))
+```
+
+> TODO: instead of there being a `GlobalLoc`, `GlobalDef` should change its
+>       meaning to be that `GlobalLoc` from the L6 onwards. However, this is
+>       currently not possible, as some IL producers (mis-)use the globals for
+>       passing constants with external meaning to the VM
+
+They may be used as a `Copy` source and in paths, much like locals.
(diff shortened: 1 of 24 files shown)


zerbina:
Next up is using the new global locations in `skully` and `source2il`, which will unblock a few tasks (notably `Ptr` support), as well as making the `total_memory` and `stack_start` memory configuration variables obsolete.

languages/lang6.md
```diff
@@ -0,0 +1,35 @@
...
+dataInit -= (Data <int> <int>)
+          | (Data <int> (StringVal <string>))
+dataInit += (Data <type>)
+          | (Data <type> (StringVal <string>))
+```
+
+> TODO: instead of there being a `GlobalLoc`, `GlobalDef` should change its
+>       meaning to be that `GlobalLoc` from the L6 onwards. However, this is
+>       currently not possible, as some IL producers (mis-)use the globals for
+>       passing constants with external meaning to the VM
```
saem, line 25: Would this imply some sort of extern facility, where we can mark certain locations as need to be kept stable for the host?

saem (approved):

languages/lang6.md
zerbina: It wouldn't, no.

Exportable constants are only needed for the host memory configuration variables, two of which (i.e., `total_memory` and `stack_start`) are now obsolete. The stack size configuration variable (`stack_size`) will be replaced with some dedicated language/module feature in the future, but I haven't fully figured out the details yet.

In other words, exportable constants in their current form are moribund. Once they're all gone, `GlobaLoc` can be removed again, with `GlobalDef` taking over its meaning.

PR #173, opened 2025-05-19
skully: fix various bugs

zerbina:
## Summary

Fix various bugs with `skully`. For the most part, they were about
incorrect code being emitted, resulting in misbehaving programs.

## Details

* fix size/alignment of embedded records (has no observable effect on
  compilation, at the moment)
* fix `nil` not being handled in constants, causing a crash
* fix incorrect code being emitted for closure equality tests
* fix pointer-location default initialization using an integer zero
* fix zero being used instead of `Nil` in various places
* fix wrong conversion operators being used for integer conversion
* fix integer extension conversion operators not being emitted
* fix `ord` not being translated to a conversion
* fix code emitted for `mDestroy` accessing the wrong field
* fix wrong code being emitted for `cast` where at least one operand is
  a float or imported type

### DIFF
--- a/skully/backend.nim
+++ b/skully/backend.nim
@@ -2368,8 +2368,14 @@ proc translateExpr(c; env: var MirEnv, tree; n; dest: Expr, stmts) =
           c.genSetElem(env, tree, it, tree[n].typ, bu)
         c.genIncl(env, dest, e, stmts)
   of mnkCast:
-    let dst = env.types.canonical(tree[n].typ)
-    let src = env.types.canonical(tree[n, 0].typ)
+    proc canon(env: TypeEnv, id: TypeId): TypeId =
+      ## Skips to the non-imported canonical type.
+      result = env.canonical(id)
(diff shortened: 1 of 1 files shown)


zerbina:
Most of these problems were found as part of #146, thanks to the stronger type system (when compared to the VM) of LLVM.

Most of the issues would have also been caught by a type checking pass that `skully` runs on its output, at least when used in conjunction with building `phy` with `skully`. As evidenced by the amount of issues, just running a `skully`-compiled `phy` does not provide enough coverage.

Testing `skully` via unit-like tests is hard, due the MIR + magics having a very large surface area -- there are also no NimSkull-provided facilities for constructing standalone MIR modules (which would be a requirement for testing). Running NimSkull's test suite with `skully` is also not possible at the moment, due to `skully` not using the `nim` CLI.

Ultimately, `skully` shouldn't have do the translation work it has do now; most of it should be done by the NimSkull compiler. I'm still looking into ways to upstream parts of `skully`.

saem (approved):

PR #176, opened 2025-05-26
source2il: fix crash when copying records

zerbina:
## Summary

Fix the compiler crashing for code where a record having a seq member
is copied.

## Details

The implementation for non-trivially-copyable records was missing,
causing the `unreachable` to trigger.

In order to prevent similar bugs going forward, the case statement for
producing the copy implementation is made exhaustive, using the new
`AllTypes` set. If a new aggregate type is added in the future, a
compile-time error will be reported if the new type is not handled.


zerbina:
The main takeaway here is, in my opinion, that sets and exhaustive case statements need to be used more often. Although not being a new insight, it's not applied as often as it should.

I think the only somewhat reliable way to catch this kind of issue is by using property-based testing. A runnable language definition already exists, so this shouldn't be too hard to do.

saem (approved):

PR #177, opened 2025-05-27
source2il: fix crashes with erroneous code

zerbina:
## Summary

Fix the compiler crashing for some instances of semantically
invalid code.

## Details

Error types reached into `typeToString`, triggering the `unreachable`
assertion and thus crashing the compiler.

* change `typeToString` to render errors as `?`. There are valid cases
  for an error type being rendered, such as when it's part of a
  composite type
* improve error recovery for tuple analysis. An element being an error
  no longer short-circuits analysis by returning an error
* use a single procedure for handling type unification (`unify`). This
  fixes unnecessary follow-up errors being reported for an `If` where
  unification fails due to one type being an error
* don't report unnecessary follow-up errors in `fitExprStrict`

Tests are added to make sure code that previously crashed the compiler
no longer does. The change to `typeToString` itself is enough to fix the
crash -- the other changes are to not report useless diagnostics.

---

## Notes For Reviewers
* found while using the language for some real-world code

### DIFF
--- a/tests/errors/terror_in_call_argument.test
+++ b/tests/errors/terror_in_call_argument.test
@@ -1,0 +1,9 @@
+discard """
+  description: "Either the if or else branch has an error"
+  reject: true
+"""
+(Module
+  (ProcDecl (Ident "test") (UnitTy) (Params (ParamDecl (Ident "x") (IntTy)))
+    (Return))
+  (ProcDecl (Ident "main") (UnitTy) (Params)
+    (Call (Ident "test") (Ident "missing"))))
--- a/tests/errors/terror_in_call_argument.test
+++ b/tests/errors/terror_in_call_argument.test
@@ -1,5 +1,5 @@
 discard """
-  description: "Either the if or else branch has an error"
+  description: "A call argument has an error"
   reject: true
 """
 (Module
--- a/tests/errors/terror_in_if.test
+++ b/tests/errors/terror_in_if.test
@@ -1,0 +1,7 @@
+discard """
+  description: "Either the if or else branch has an error"
+  reject: true
+"""
+(If (Ident "true")
+  (Ident "missing")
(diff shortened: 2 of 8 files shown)


tests/errors/terror_in_tuple_cons.test
```diff
@@ -0,0 +1,5 @@
+discard """
+  description: "Either the if or else branch has an error"
```
saem, line 2:
```suggestion
  description: "Error in a field of a tuple"
```

I believe this is what you might have meant

tests/errors/terror_in_if.test
```diff
@@ -0,0 +1,7 @@
+discard """
+  description: "Either the if or else branch has an error"
+  reject: true
+"""
+(If (Ident "true")
+  (Ident "missing")
+  (IntVal 10))
```
saem, line 7: Given the recent issues that were surface, it's probably a good idea to make the testing exhausive by putting an error in the else branch as a separate test.

The other test can be a simple reversal:

```nim
iscard """
  description: "Either the if or else branch has an error"
  reject: true
"""
(If (Ident "true")
  (IntVal 10))
  (Ident "missing")
```

tests/errors/terror_in_call_argument.test
```diff
@@ -0,0 +1,9 @@
+discard """
+  description: "Either the if or else branch has an error"
```
saem, line 2:
```suggestion
  description: "Error in a call argument"
```

I believe this is what you might have meant

saem (approved):
A few small suggestions wrt testing

tests/errors/terror_in_tuple_cons.test
zerbina: Oh, I forgot to update the description after duplicating the if test, thanks.

PR #179, opened 2025-05-30
implement and use a proper nanopass framework

zerbina:
While the initial goal was to use a nanopass-based architecture for the
compiler, this is not really the case, at present. The characteristic of
nanopasses are that:
1. they focus on a small transformation or analysis task
2. they contain no traversal boilerplate
2. the input and output language are well-defined

The current pass architecture only insufficiently matches the above.
While some tricks are used to keep the traversal logic small, it's
still present.

The goal with this PR is to implement a nanopass framework and write a
compiler for the source language with it.

Passes and languages will become much easier to write. Compared to the
current state, some notable improvements are that:
* it is ensured at *compile time* (instead of at *run-time*) that
  constructed AST adheres to the IL's grammar
* AST can embed **arbitrary non-AST data**
* traversal logic in passes is auto-generated in most cases
* languages are defined and available within the host language (i.e.,
  NimSkull)

This makes the passtool obsolete, though it will have to be kept around
for now, as - in order to keep the scope of this PR smaller - skully
is going to continue using the previous pass pipeline.

---

## To-Do
* [x] add simple source location tracking support to the nanopass framework
* [x] add record type support to the nanopass framework
* [x] address all important remaining inline TODO comments
* [x] reject language definitions containing forms that violate internal expectations 
* [ ] finish the new nanopass-using compiler

### DIFF
--- a/passes/passes_legacy.nim
+++ b/passes/passes_legacy.nim
@@ -188,12 +188,12 @@ proc basicBlocks(ir: Lskully): L6 {.pass.} =
 
   proc target(x: src.tgt, map: Table[int64, int]): dst.tgt =
     match x:
-    of Goto(i): build dst.tgt, Goto(i(^map[i.val]))
-    of Unwind(): build dst.tgt, Unwind()
+    of Goto(i): build dst.tgt, x.info, Goto(i(^map[i.val]))
+    of Unwind(): build dst.tgt, x.info, Unwind()
 
   proc goto(x: src.go, map: Table[int64, int]): dst.go =
     match x:
-    of Goto(i): build dst.go, Goto(i(^map[i.val]))
+    of Goto(i): build dst.go, x.info, Goto(i(^map[i.val]))
 
   type BBlock = object
     isExcept: bool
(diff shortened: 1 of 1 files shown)


zerbina:
Language definition and extension works, but the pass DSL is still missing. The `passes` module contains some pass examples using a preliminary syntax, but it's possible that the syntax won't work and has to be replaced with something else.

nanopass/nanopass.nim
```diff
@@ -0,0 +1,445 @@
...
+        quote do: `p`(`q` do: `filtered`)),
+      body[0])
+  else:
+    result = nnkConstSection.newTree(
+      nnkConstDef.newTree(name,
+        newEmptyNode(),
+        quote do: `p`(`q` do: `body`)))
+
+macro defineLanguage*(name, base, body: untyped) =
+  ## Creates a language definition, extending `base`, and binds it to a const
```
saem, line 427: Is the intention, that after nanopass is implemented, to look at what it would mean to go with an extension/inheritance vs composition based API?

nanopass/nanopass.nim
```diff
@@ -0,0 +1,445 @@
...
+  computeTags(result)
+
+proc makeLanguage(body: NimNode): LangDef =
+  ## Creates a language definition from the ``defineLanguage`` DSL code.
+  body.expectMinLen 1
+  var add: seq[NimNode]
+  var def: seq[NonTerminalDef]
+
+  # second pass: process the productions
+  proc extract(n: NimNode, list: var seq[NimNode]) =
```
saem, line 317: I like the internal procs, I'm finding them quite a bit more readable than having them at the module level.

nanopass/nanopass.nim
zerbina: Yep, I want to have something that's certain to work first, which is also the reason metadata is not part of the PR.

saem (commented):
First cut at documentation, the next thing I'm going to work at is describe, hopefully via examples/graphically, how various functions decompose grammar structures, especially `parseForm`

If you want me to make any major changes in terms of docs, let me know, right now this is mostly forcing me to understand the code as I work through it and try to explain it.

nanopass/nanopass.nim
```diff
@@ -0,0 +1,445 @@
+## Implements the nanopass framework, which is a collection of macro DSLs for
+## defining intermediate languages (their syntax and grammar) and passes.
```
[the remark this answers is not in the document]
zerbina: Nice, looks good, and good call on favoring brevity. There's still the link to the paper for those curious/interested in more details.

I think it'd be a good idea to also link to the [dissertation](https://andykeep.com/pubs/dissertation.pdf) Scheme's nanopass framework is based on, given that the latter was the main inspiration for the implementation (so far).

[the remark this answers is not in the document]
saem: Updated with the dissertation link, I should be able to sort out the rest of it once I have the routines below documented a bit more.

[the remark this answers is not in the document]
saem: I've updated again, I personally think I'm missing some nuance, not necessarily in the description, but in my mental model about non-terminal form definitions: `nonterm(nt) ::= Form(...nt)`, here the lhs `nonterm` is a nonterminal form, with `nonterm` being it's name, and ranged over by `nt`. Where `Form(...nt)` is a nonterminal named form, but also a patten?

I need to sort out the confusion above, but I'll leave that for tomorrow/day after once I've reread the code some more.

zerbina:
I've pushed some cleaned-up progress on the nanopass framework, plus an updated version of the ILs/passes I'm currently using for testing.

It's important to note that the current ILs and passes only exist for testing and demonstration -- they are incomplete, incorrect, and there are some glaring problems with ordering. With my local nanopass framework, I'm able to compile *and* run `passes.nim`, where "run" here means that the code doesn't crash and produces an AST that has the correct *shape*.

What's mostly final is the general shape of the passes' body (i.e., there being zero or more processor procedures), the rest (syntax, naming, etc.) is still in-flux to varying degree.

---

One major outstanding question is how to best represent type information. Currently, types are coupled with languages via their construction forms being part of the language definitions. Type information is part of the forms via "type tags", that is, type constructions being the first element in most forms (after type checking). For example:
`(If (BoolTy) (...) (...) (...))`, or `(TupleCons (TupleTy (IntTy) (FloatTy)) 1 1.0)`.

This has some major downsides:
* **in-memory size of ASTs.** Type construction sub-trees are potentially duplicated many times. Not only does this mean increased space taken up by the total AST, it also means more processing work for passes (more AST to traverse, copy, etc.).
* **types cannot be referenced from the outside.** Symbols may want to refer to types too, but cannot, at the moment. There are some ways to make this work (even without extra nanopass framework support), but I don't particularly like them.
* **recursive types aren't supported directly.** To be able to represent recursive types, some form of type symbols (i.e., types bound to identifiers) is needed, which adds complexity to passes that need to inspect types (but there might be ways around this).

There are multiple possible approaches/solutions here.

For one, the same strategy as the one the legacy ILs use can be used. That is, keeping a list of types on the module and referring to them via an index (effectively a symbolic reference). This requires manually handling book-keeping and lookup, however, which is cumbersome and prone to mistakes and I'd rather not go down that route.

Another approach could be to move type constructions to their own set of languages. The "main" languages would then refer to types via name terminals (this is a strategy I'm experimenting with as part of some NimSkull work, for what it's worth), or by embedding the types directly as terminals. Type lowering would need to happen in separate passes, which is troublesome when embedding the types directly as terminals, as the relevant "main" language passes and their associated type lowering pass would need to happen atomically, and I'm not sure how that could be made to work in practice.

Yet another approach could be to have the nanopass framework transparently "deduplicate" AST fragments, using node references internally (when deduplicating). While this would only address the size concern, having this feature might be a good idea in general.

nanopass/manual.rst
```diff
@@ -0,0 +1,213 @@
...
+
+.. code-block:: nim
+    :test: "nim c $1"
+    :status: 1
+
+  import nanopass/nanopass
+
+  defineLanguage L0:
+    i(int)
+    integer(i) ::= i # error: 'i' name already in use
```
saem: Doesn't have to change, but this feels like the error message should instead be more like "attempting to range over a metavariable, provide a type instead", unless I'm forgetting something (highly likely). I get the names exist in the same space, but I thought they were still distinct in terms of type (ranging over something metavar vs type name).

I'll be doing some rereading to remind myself of these bits, so I might update the comment in a few days.

saem: ok, I see where I'm confused and how things went off the rails from there, closing this comment.

nanopass/manual.rst
```diff
@@ -0,0 +1,213 @@
...
+The body consists of a sequence of terminal and non-terminal definitions.
+
+.. code-block:: nim
+    :test: "nim c $1"
+
+  import nanopass/nanopass
+
+  defineLanguage L0:
+    i(int)         # definition of a terminal
+    expr(e) ::= i  # definition of a non-terminal
```
saem: I think people might trip over this:

```
<new_metavar>(<terminal_type>)
# vs
<nonterminal_type>(<new_metavar>) ::= ...
```

The former introduces a new metavar over the thing inside the parens, so it's declares a name and binds it to what's inside the parens. While the latter  is a metavar in the parens introduced to range over the thing outside the parens, which itself is introducing a new name, and binding that to the production pattern on the right.

The syntactic flip between the metavar inside vs outside the parens is weird at first, so I'm thinking about how to better explain it.

---

I personally don't have a better proposed syntax, and whatever the change I think it'd be pretty big. So I'm trying to sort out how to make it more approachable from a documentation perspective.

zerbina: I've got used to it, but the same sub-syntax having flipped meaning isn't great, yea.

I think it would make sense to swap the operands for terminal definitions, such that `i(int)` becomes `int(i)`. This would be much easier to teach and explain:
* a non-terminal definition introduces a new type (with the name given by the name outside the parens)
* a terminal definition merely makes a pre-existing type (given by the name outside the parens) available in the grammar

In both cases, the type is now the one outside the parenthesis, and terminals could also have more than one metavar.

PR #181, opened 2025-08-10
interpreter: improve plug pattern matching

zerbina:
## Summary

Improve meta-language interpreter's plug pattern matching, making holes
following repetitions in lists work as expected and preparing for
further improvements.

## Details

Instead of treating hole patterns as matching everything and then
handling the match between the captured term and plugged-with pattern
once the total match has been successful, the plugged-with pattern is
used in place of the hole.

Passing values "downwards" during pattern matching is not supported, so
instead of passing the plugged-with pattern towards the hole matcher,
a `matches` encountering a hole returns the special `Match` instance
(using the `hole` kind) that contains the current continuation to the
closest handler, which then handles the hole match appropriately.

To support multi-result patterns in a more generally applicable way,
successful matches now store an optional continuation representing a
not-yet-tried match thread.

Together this makes the following pattern usages work (without an
external match driver):
1. holes following repetitions in lists
2. using a context name as a pattern in contexts
3. plug patterns part of contexts
4. holes appearing in repetitions

Only number 1 actually affects the meta language, as 2 was previously
implemented via an external match driver and both 2 and 3 are either
forbidden or already handled/inlined by the type checker.

Finally, some tests are added directly for the pattern matcher to make
sure the changes work.

### DIFF
--- a/tests/unittest/tinterpreter_matching.nim
+++ b/tests/unittest/tinterpreter_matching.nim
@@ -1,0 +1,197 @@
+discard """
+  description: "Tests for the pattern matching of the meta-language interpreter"
+"""
+
+import std/[importutils, tables]
+import spec/cps
+import spec/interpreter {.all.}
+import spec/types except Node
+
+# some procs/types to make writing tests easier
+
+type Node = types.Node[TypeId]
+
+template sym(n: string): Node = Node(kind: nkSymbol, sym: n)
+template hole(): Node = Node(kind: nkHole)
+template ctxRef(i: int): Node = Node(kind: nkContext, id: i)
+template vid(i: int): Node = Node(kind: nkVar, id: i)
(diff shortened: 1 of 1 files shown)


zerbina:
Beyond actually working like one would expect it to now, the new implementation is also a bit easier to follow, in my opinion. The `interpretMatch` was largely unintelligible previously, handling the convoluted driver logic for the hole pattern matching. Now, `interpretMatch` only needs to "iterate" through the `alt` chain of the match.

The actual pattern matcher got a little more complex, but most of this is due to the boilerplate required for the continuation creation. Adding `suspend` to NimSkull was - in part - meant to address this, the idea being it making writing some sugar for replacing the manual continuation creation possible. Unfortunately, `suspend` - at least in its current form - is not suitable for this task, as parameters cannot be used across suspension points, which is something the pattern matching implementation does a lot.

PR #191, opened 2026-03-13
improve VM module validation error messages

zerbina:
## Summary

Improve the user-facing error messages for VM module validation errors,
by generally providing more context to make figuring out where the
error originated from easier (and sometime possible in the first place).

## Details

* use simple structured data for representing the errors returned by VM
  module validation, to allow for off-loading rendering to the receiver
  * except for operand stack mismatch errors, most error messages are
    still rendered by the validator 
* reword some error messages rendered by the validator to be more
  actionable and provide more context
* move instruction rendering in `disasm` into its own procedure
* implement pretty rendering for errors in the new `vmerrors` module
* use pretty rendering in all places where validation errors
  are displayed

### Example

Given:

```
.type typ () -> void
.start typ main
  LdImmInt 0
  LdImmFloat 1
  AddInt
  Ret
.end
```

the displayed error message is now:

```
Error: in procedure 0:
  
    ...
    LdImmFloat 1
  > AddInt
    Ret

expected operand stack with shape [..., int, int], but got [...int, float]
```

instead of:

```
Error: at 2: unexpected type
```

---

## Notes For Reviewers

* this is a long overdue change, as the error messages were previously next to useless


saem (approved):
One clarification/possible fix, but otherwise looks good; the messages are much improved.

